PR #4, opened 2023-12-21
Update README.md

fdiehr:


cutoffthetop:
will harmonize boilerplate here https://github.com/robert-koch-institut/mex-editor/pull/6

PR #108, opened 2024-07-16
feature/mx-1615 setup authentication

cutoffthetop:
# PR Context
This PR adds basic authentication against a local user config, as a stop-gap for a centralized IAM.

# Added
- add login component and form
- add user menu in navbar with logout button
- setup authentication against locally configured users

# Changes
- re-organize state classes into individual modules

# Removed
- remove unused `mex.editor.transform.to_primitive`


esinsj (commented):
I could start the backend and the editor. I copied the URL in firefox and it showed the login-screen, But I couldn't log in with the new credentials. Only the old credentials, which I still had in my .env worked.

Suggestion: It would be great to have the command on how to shut down the running editor and backend in the README.md.

cutoffthetop:
> I couldn't log in with the new credentials. Only the old credentials, which I still had in my .env worked.

not sure what you mean with _the new_ credentials, but this is desired behaviour, it should validate against the ones that are found in your local .env file

> It would be great to have the command on how to shut down the running editor

yeah good point. instead of the readme, I added a log message to the startup procedure though; see 0ee3326

mr-kamran-ali (commented):
Looks good to me. this criterion is not met yet: is there a ticket for this? 
- ld** auth in apache

mr-kamran-ali (approved):

PR #117, opened 2024-07-25
feature/mx-1648 edit view

cutoffthetop:
# PR Context
- introduces new testing library [playwright](https://playwright.dev/python/) for browser tests
- see [/actions/runs/10146965614](https://github.com/robert-koch-institut/mex-editor/actions/runs/10146965614) for screenshots of integration tests running in CI
- to start tests locally, start neo4j, the latest mex-backend and the mex-editor

# Added

- implement a basic item view which will be expanded into the editing mode
- add model config YAML to store which fields to use for titles and previews
- add basic rendering functions for arbitrary models and types
- configure CI integration tests running against real backend and database
- implement integration tests for all four major pages in the editor

# Changes

- rework navigation bar to work with states instead of passing through literals
- update layout and styling on login form, merge page stub and search view


mex/editor/edit/main.py
```diff
@@ -1,12 +1,64 @@
...
+def editable_field(model: EditableField) -> rx.Component:
+    """Return a horizontal grid of cards for editing one field."""
+    return rx.hstack(
+        rx.card(
+            model.name,
+            style={"width": "15vw"},
+        ),
+        rx.foreach(
+            model.values,
+            editable_primary_source,
```
mr-kamran-ali, line 40: Why is this looping over `editable_primary_source` and not `editable_field`. I think names here might be a little confusing.

mex/editor/models.yaml
```diff
@@ -0,0 +1,46 @@
+AccessPlatform:
+  title: title
+  preview:
+    - alternativeTitle
+    - unitInCharge
+Activity:
+  title: title
+  preview:
```
mr-kamran-ali, line 8: so this model for now represents the fields that can be edited? and later on we can just append more fields under `preview` if we want to edit more fields?

pyproject.toml
```diff
@@ -10,18 +10,19 @@ requires-python = "<3.12,>=3.11"
...
     "babel==2.15.0",
     "mex-common@git+https://github.com/robert-koch-institut/mex-common.git@0.32.0",
+    "pyyaml==6.0.1",
     "reflex==0.5.7",
     "uvicorn==0.20.0",
 ]
 optional-dependencies.dev = [
     "ipdb==0.13.13",
     "mypy==1.11.0",
-    "pytest-cov==5.0.0",
```
mr-kamran-ali, line 19: pytest coverage does play nice with playwright?

tests/login/test_main.py
```diff
@@ -1,6 +1,20 @@
...
+        writer_user_credentials[0],
+    )
+    page.get_by_placeholder("Password").fill(
+        writer_user_credentials[1].get_secret_value(),
+    )
+    page.screenshot(path="tests_login_test_main-test_login-on-load.jpeg")
+
+    page.get_by_text("Log in").click()
+    expect(page.get_by_test_id("nav-bar")).to_be_visible()
+    page.screenshot(path="tests_login_test_main-test_login-after-login.jpeg")
```
mr-kamran-ali, line 20: what usecase did you have in mind for these screenshots? are they for human verification at the end? or for visual logging?

mr-kamran-ali (commented):
`/edit/test_main.py::test_index` and `/search/test_main.py::test_index` tests are failing with 422 Client error(`Unprocessable Entity for url: http://localhost:8080/v0/ingest`). This is coming from `mex-backend` that I am running locally as I see in the `mex-backend` stdout. 

Other than few clarifications, looks good.

tests/login/test_main.py
cutoffthetop: mainly for debugging a test. if you see a test failing, you can look at a screenshot to figure out what the browser would look like, and why the assertion didn't pass.

also works for CI, which i find really helpful. see the **Artifacts** section here https://github.com/robert-koch-institut/mex-editor/actions/runs/10146965614

pyproject.toml
cutoffthetop: sadly not that easily because the editor runs in a different process. but i'm also not sure how helpful it would be for this sort of visual / browser-based testing

mex/editor/models.yaml
cutoffthetop: no, it's just how the configuration for how to display each model type in the frontend for example when you do a search or have a link field.

see https://github.com/robert-koch-institut/mex-editor/pull/117/files#diff-241def25481c545bfed47d68895cbbeb903b70358b8c52b189f341491ca9d1e2R15

because not all models have a "title" and "description" field we could use, each model type will look a little different.

mex/editor/edit/main.py
cutoffthetop: it's because of how the edit UI is structured:
- each edit page has a list of field sections (depending on which entity type you are editing)
  - each field section has a list of primary-sources (depending on how many extracted items will be merged / have the same stable target id)
    - each primary-source section has a list of concrete values (because even from one primary source, you might extract multiple values)

i changed the name of the "values" attribute here ffba2e5. hopefully that's a bit clearer

cutoffthetop:
> tests are failing with 422 Client error

oh good catch, i needed to update mex-common to 0.33 to include https://github.com/robert-koch-institut/mex-common/pull/245 -> done here: 5d39772

mex/editor/models.yaml
mr-kamran-ali: got it. thanks for the explanation.

tests/login/test_main.py
mr-kamran-ali: yeah, just checked the artifacts. didnt realize they'll be available in github CI. this is cool.

mr-kamran-ali (approved):
Looks good. nice work ❤️

PR #151, opened 2024-09-10
feature/set fastapi, starlette and pydantic versions

cutoffthetop:
# Changes
- set fastapi, starlette and pydantic versions explicitly and update mex-common


erichesse (approved):
Looks good!

PR #192, opened 2024-11-11
feature/mx-1649 prevent and subtract

cutoffthetop:
# PR Context
- for local testing, needs a backend started with debug=True and identity_provider=graph
- removed localization of temporals because those don't convert back to rules that easily. let's wait until localization becomes a hard requirement

# Added
- add toggles for preventive and subtractive rules
- add functionality to edit component for submitting rules
- add utility function to escalate errors to all consoles
- temporarily add BackendIdentityProvider (stop-gap MX-1763)

# Changes
- rename FixedX and EditableX classes to EditorX for consistency
 
# Removed

- drop dev-dependency to mex-backend, use the flush endpoint instead
- temporarily removed localization of temporals entity output


tests/conftest.py
```diff
@@ -49,6 +47,20 @@ def settings() -> EditorSettings:
...
+    """Ensure the identifier provider is set correctly for unit and int tests."""
+    # TODO(ND): clean this up after MX-1708
+    for settings in (BaseSettings.get(), EditorSettings.get()):
+        if is_integration_test:
+            monkeypatch.setitem(settings.model_config, "validate_assignment", False)
+            monkeypatch.setattr(
+                settings, "identity_provider", EditorIdentityProvider.BACKEND
+            )
+        else:
+            monkeypatch.setattr(settings, "identity_provider", IdentityProvider.MEMORY)
```
rababerladuseladim: There should only be one settings instance. I am however surprised that this does not throw an error in mex common...

```suggestion
    settings = EditorSettings.get():    
    if is_integration_test:
        monkeypatch.setitem(settings.model_config, "validate_assignment", False)
        monkeypatch.setattr(
            settings, "identity_provider", EditorIdentityProvider.BACKEND
        )
    else:
        monkeypatch.setattr(settings, "identity_provider", IdentityProvider.MEMORY)
```

tests/conftest.py
```diff
@@ -211,7 +197,10 @@ def dummy_data() -> list[AnyExtractedModel]:
 
 
 @pytest.fixture
-def load_dummy_data(dummy_data: list[AnyExtractedModel]) -> list[AnyExtractedModel]:
+def load_dummy_data(
+    dummy_data: list[AnyExtractedModel],
+    flush_graph_database: None,  # noqa: ARG001
```
rababerladuseladim, line 201: Why not `pytest.usefixture`?

mex/editor/edit/main.py
```diff
@@ -1,94 +1,185 @@
...
+def editor_value_switch(
+    field_name: str,
+    primary_source: str | None,
+    value: EditorValue,
+    index: int,
+) -> rx.Component:
+    """Return a switch for toggling subtractive rules."""
+    return rx.switch(
+        checked=value.enabled,
+        on_change=lambda enabled: cast(EditState, EditState).toggle_field_value(
```
rababerladuseladim, line 20: why is casting required here?

mex/editor/transform.py
```diff
@@ -1,125 +1,113 @@
...
-            external=False,
+        return EditorValue(
+            text=str(value),
+            badge=value.precision.value,
         )
-    if value is not None:
-        return FixedValue(
+    if isinstance(value, str):
+        return EditorValue(
             text=str(value),
```
rababerladuseladim, line 59:
```suggestion
            text=value,
```

mex/editor/edit/transform.py
```diff
@@ -1,39 +1,198 @@
...
+    """Given a model, a field and an optional subtractive rule, create editor values."""
+    model_values = ensure_list(getattr(model, field_name))
+    editor_values = []
+    for model_value in model_values:
+        editor_value = transform_value(model_value)
+        editor_value.enabled = not (
+            subtractive
+            and field_name in MERGEABLE_FIELDS_BY_CLASS_NAME[subtractive.entityType]
+            and model_value in getattr(subtractive, field_name)
+        )
```
rababerladuseladim: I struggled to understand when this condition is met. Could you introduce a speaking variable for it? E.g. `model_value_is_merbeable_and_subtracted_by_rule`?

mex/editor/edit/transform.py
```diff
@@ -1,39 +1,198 @@
...
+                    identifier=primary_source_id,
+                    editor_values=_transform_model_values_to_editor_values(
+                        model, field_name, subtractive
+                    ),
+                    enabled=not (
+                        preventive
+                        and field_name
+                        in MERGEABLE_FIELDS_BY_CLASS_NAME[preventive.entityType]
+                        and primary_source_id in getattr(preventive, field_name)
+                    ),
```
rababerladuseladim: same for this. `field_is_mergeable_and_prevented_by_rule`

mex/editor/identity.py
```diff
@@ -0,0 +1,66 @@
...
+        """Set the backend API key to all session headers."""
+        settings = EditorSettings.get()
+        self.session.headers["X-API-Key"] = settings.backend_api_key.get_secret_value()
+
+    def _set_url(self) -> None:
+        """Set the backend api url with the version path."""
+        settings = EditorSettings.get()
+        self.url = urljoin(str(settings.backend_api_url), self.API_VERSION)
+
+    @cache  # noqa: B019
```
rababerladuseladim: Why is `B019` safe to ignore? Because BackendIdentityProvider is a singleton? If yes, could you add that to the comment?

tests/edit/test_main.py
```diff
@@ -102,21 +117,102 @@ def test_edit_link(edit_page: Page) -> None:
...
+@pytest.mark.parametrize(
+    ("switch_id"),
+    [
+        (r"switch-abstract-{primary_source_id}"),
+        (r"switch-abstract-{primary_source_id}-1"),
+    ],
+    ids=["toggle primary source", "toggle value"],
+)
+@pytest.mark.integration
+def test_edit_page_switch_roundtrip(
```
rababerladuseladim, line 158: wow, what a journey O.O

rababerladuseladim (changes requested):
wow...

mex/editor/identity.py
cutoffthetop: done 588f5fa

mex/editor/transform.py
cutoffthetop: it might still be a `str` subclass after the `isinstance` check, like `Email`

mex/editor/edit/main.py
cutoffthetop: mypy complains about `toggle_field_value` being an instance method, not a class method. meanwhile, reflex is doing some dark singleton magic with States, so that this method is actually being called on the singleton instance during runtime

tests/conftest.py
cutoffthetop: that doesn't work and would fail https://docs.astral.sh/ruff/rules/pytest-erroneous-use-fixtures-on-fixture

> Note this mark [usefixture] has no effect in fixture functions

see, https://docs.pytest.org/en/8.3.x/how-to/fixtures.html

mex/editor/edit/transform.py
cutoffthetop: i simplified it a bit (the subtractive argument didn't actually have to be optional), and added some comments: a66e32d

mex/editor/edit/transform.py
cutoffthetop: see also a66e32d

tests/conftest.py
cutoffthetop: you're right. both `BaseSettings.get()` and `EditorSettings.get()` returned the same `EditorSettings` instance, and the loop would just do the same thing twice 🤦

rababerladuseladim (approved):

PR #213, opened 2024-12-12
redirect to original url after login

ZehraVictoria:
# Added

- Redirect user to original URL after successful login

### DIFF
--- a/mex/editor/state.py
+++ b/mex/editor/state.py
@@ -11,6 +11,7 @@ class State(rx.State):
     """The base state for the app."""
 
     user: User | None = None
+    current_url: str | None = None
     nav_items: list[NavItem] = [
         NavItem(title="Search", href_template=r"/"),
         NavItem(title="Edit", href_template=r"/item/{item_id}/"),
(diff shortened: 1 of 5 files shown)


mex/editor/state.py
```diff
@@ -11,6 +11,7 @@ class State(rx.State):
     """The base state for the app."""
 
     user: User | None = None
+    current_url: str | None = None
```
cutoffthetop: could we find a better name for this attribute? 🤔 

it will only store the "current_url" before you login. after the login redirecting is done and the user is clicking around in the app, this attribute will not be updated. and that might be confusing to future developers if they try to use this for other purposes..

maybe something like `target_path_after_login`? a little clunky, but at least no one will try to use this thinking it will stay in sync with the browser's url bar

cutoffthetop (changes requested):
nice and straightforward solution! i have just one _naming-things-gripe_

mex/editor/state.py
ZehraVictoria: Yes, good point - Changed the name accordingly :)

cutoffthetop (approved):
perfect! merge away 🛩️

PR #248, opened 2025-02-05
feature/mx-1741 basic field editing

cutoffthetop:
### Added

- add color mode toggle to login screen
- added new/remove buttons for additive rules for string fields
- add event handlers for adding/removing additive value input cards
- added text input field for email, string and temporal fields
- add event handler for setting string input values

### Changes

- combine EditState.editor_fields and EditState.fields
- update transformation logic to handle strings from additive rules


mex/editor/edit/transform.py
```diff
@@ -70,9 +72,48 @@ def _transform_model_values_to_editor_values(
...
+
+def _create_editor_primary_source(  # noqa: PLR0913
+    primary_source_name: EditorValue,
+    primary_source_id: MergedPrimarySourceIdentifier,
+    editor_values: list[EditorValue],
+    field_name: str,
+    preventive: AnyPreventiveModel,
+    input_config: InputConfig | None,
+) -> EditorPrimarySource:
+    """Create a new editor primary source from the given parameters."""
```
rababerladuseladim: `editor primary source` could also mean "Primary Source for elements originating in the editor".

```suggestion
    """Create a new primary source from the given parameters."""
```

mex/editor/edit/main.py
```diff
@@ -67,35 +169,82 @@ def primary_source_name(
         rx.hstack(
             render_value(model.name),
             rx.cond(
-                cast(rx.vars.ArrayVar, EditState.editor_fields).contains(field_name),
+                ~cast(rx.vars.BooleanVar, model.input_config),
```
rababerladuseladim: Why is this a boolean? Why can you apply `~` to it? What happens here???

mex/editor/edit/transform.py
```diff
@@ -81,72 +122,127 @@ def _transform_model_to_editor_primary_source(
...
+                field_name,
+                subtractive,
+            )
+            if isinstance(model, AnyAdditiveModel):
+                if not (
+                    input_config := _transform_model_to_additive_input_config(
+                        field_name,
+                        model.entityType,
+                    )
+                ):
```
rababerladuseladim: I find this easier to comprehend.
```suggestion
                input_config = _transform_model_to_additive_input_config(
                        field_name,
                        model.entityType,
                    )
                if input_config is None:
```

rababerladuseladim (changes requested):
Oh boi, this is a lot. I am far from understanding everything but tried to review as best as I could. 

I stumbled upon one UI issue:
- primary source switches are sometimes not visible, e.g.: http://{editor-url}/item/27WuXh9ge9voadrk9y9t0/

mex/editor/edit/main.py
cutoffthetop: `~` is a var operation https://reflex.dev/docs/vars/var-operations/#and,-or-and-not 
the casting to boolean was technically not correct, true. i changed it to `ObjectVar`

cutoffthetop:
> I stumbled upon one UI issue: primary source switches are sometimes not visible

fixed in https://github.com/robert-koch-institut/mex-editor/pull/248/commits/b0518311488ccbc1fa930e7348be23f81cd3861d

rababerladuseladim (approved):

mex/editor/edit/main.py
rababerladuseladim: I do not understand how you can apply the `~` operator to a variable that is not a bool. `input_config` is `InputConfig | None`, neither of which is a bool. Could you elaborate?

cutoffthetop: it's not _actually_ `InputConfig | None` though. it's an instance of a dynamic subclass of `reflex.vars.base.Var[InputConfig | None]`.

and the `~` does not _actually_ evaluate in python. reflex just overloaded the `__invert__` to return an instruction that is translated to javascript and evaluated by the browser at runtime.

> Var operations transform the placeholder representation of the value on the frontend and provide a way to perform basic operations on the Var without having to define a computed var.

see https://reflex.dev/docs/vars/var-operations/#var-operations

PR #263, opened 2025-02-27
feature/mx-1679 import organizations from wikidata

ZehraVictoria:
### Added
- import button to ingest organizations from wikidata to backend

### DIFF
--- a/mex/editor/aux_search/main.py
+++ b/mex/editor/aux_search/main.py
@@ -25,16 +25,13 @@ def expand_properties_button(result: AuxResult, index: int) -> rx.Component:
 def import_button(result: AuxResult, index: int) -> rx.Component:
     """Render a button to import the aux search result to the MEx backend."""
     return rx.cond(
-        result.import_button_disabled,
+        result.show_import_button,
         rx.button(
             "Import",
             align="end",
-            disabled=True,
         ),
         rx.button(
-            "Import",
-            on_click=lambda: AuxState.import_result(index),  # type: ignore[missing-parameter]
-            align="end",
+            "Import", on_click=AuxState.import_result(index), align="end", disabled=True
         ),
     )
 
--- a/mex/editor/aux_search/main.py
+++ b/mex/editor/aux_search/main.py
@@ -22,13 +22,20 @@ def expand_properties_button(result: AuxResult, index: int) -> rx.Component:
     )
 
 
-def import_button(index: int) -> rx.Component:
+def import_button(result: AuxResult, index: int) -> rx.Component:
     """Render a button to import the aux search result to the MEx backend."""
-    return rx.button(
-        "Import",
-        on_click=lambda: AuxState.import_result(index),
-        align="end",
-        disabled=True,
+    return rx.cond(
+        result.import_button_disabled,
+        rx.button(
+            "Import",
+            align="end",
+            disabled=True,
+        ),
+        rx.button(
+            "Import",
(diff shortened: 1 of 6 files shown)


mex/editor/aux_search/models.py
```diff
@@ -11,3 +11,4 @@ class AuxResult(rx.Base):
     preview: list[EditorValue]
     show_properties: bool
     all_properties: list[EditorValue]
+    import_button_disabled: bool
```
cutoffthetop: formulating booleans in a "positive way" makes it easier to read statements that use them. this way it is more harmonic with `show_properties`, too.

```suggestion
    show_import_button: bool
```

mex/editor/aux_search/main.py
```diff
@@ -22,13 +22,20 @@ def expand_properties_button(result: AuxResult, index: int) -> rx.Component:
...
+    return rx.cond(
+        result.import_button_disabled,
+        rx.button(
+            "Import",
+            align="end",
+            disabled=True,
+        ),
+        rx.button(
+            "Import",
+            on_click=lambda: AuxState.import_result(index),  # type: ignore[missing-parameter]
```
cutoffthetop:
```suggestion
            on_click=AuxState.import_result(index),
```

this wouldn't work in plain python, but reflex has some magic going on to simplify the syntax of passing arguments to event handlers, similar to [functools.partial](https://docs.python.org/3.11/library/functools.html#functools.partial).

see https://reflex.dev/docs/events/event-arguments/

mex/editor/aux_search/state.py
```diff
@@ -81,12 +81,13 @@ def import_result(self, index: int) -> Generator[EventSpec | None, None, None]:
         """Import the selected result to MEx backend."""
         connector = BackendApiConnector.get()
         try:
-            connector.ingest([self.results_extracted[index]])
+            connector.ingest([self.results_extracted[index].model_copy(deep=True)])
```
cutoffthetop: maybe something like this would make it more obvious, what is happening. you are de-referencing the model from the proxy object, you don't actually need a deep copy.

```python
from reflex.state import serialize_mutable_proxy

model = serialize_mutable_proxy(self._results_extracted[index])
connector.ingest([model])
```

cutoffthetop (changes requested):
very nice! just two readability-suggestions from my side.

phew, and wikidata api is slooow 🥱

mex/editor/aux_search/state.py
ZehraVictoria: fixed in [7d57a33](https://github.com/robert-koch-institut/mex-editor/pull/263/commits/7d57a339feddd3d4ea0fcc55fad86ad36e7f8514)

mex/editor/aux_search/main.py
ZehraVictoria: fixed in [7d57a33](https://github.com/robert-koch-institut/mex-editor/pull/263/commits/7d57a339feddd3d4ea0fcc55fad86ad36e7f8514)

mex/editor/aux_search/models.py
ZehraVictoria: Yes, makes sense :)
fixed in [7d57a33](https://github.com/robert-koch-institut/mex-editor/pull/263/commits/7d57a339feddd3d4ea0fcc55fad86ad36e7f8514)

cutoffthetop (approved):
all issues resolved ✨ feel free to merge!

PR #269, opened 2025-03-04
feature/mx-1679 search and import from ldap

ZehraVictoria:
### PR Context

- unfortunatly both integration test behave quite unreliable locally (I think its because wikidata and the intial ldap search request take a while)
### Added
- add search and import for LDAP


mex/editor/aux_search/state.py
```diff
@@ -116,7 +121,7 @@ def search(self) -> Generator[EventSpec | None, None, None]:
         except HTTPError as exc:
             self.reset()
             yield from escalate_error(
-                "backend", "error fetching wikidata items: %s", exc.response.text
+                "backend", "error fetching items: %s", exc.response.text
```
cutoffthetop:
```suggestion
                "backend", "error fetching items", exc.response.text
```

mex/editor/aux_search/main.py
```diff
@@ -172,16 +175,51 @@ def search_results() -> rx.Component:
...
-            lambda item: rx.text(
-                item,
-                cursor="pointer",
-                size="5",
+        rx.tabs.root(
+            rx.tabs.list(
+                rx.spacer(),
+                rx.tabs.trigger("Wikidata", value="wikidata"),
+                rx.tabs.trigger("LDAP", value="ldap"),
+                rx.tabs.trigger("Orchid", value="orchid"),
```
cutoffthetop: could you move this into an attribute on the state? similar to https://github.com/robert-koch-institut/mex-editor/blob/0.12.0/mex/editor/state.py#L17

knowing about which "aux providers" we have and which one the default is, should be the job of the business layer, not the presentation layer[^1]. 

[^1]: https://en.wikipedia.org/wiki/Multitier_architecture

cutoffthetop (changes requested):
nice and simple so far, i have 1 architectural request though

mex/editor/aux_search/main.py
ZehraVictoria: Noted and applied in https://github.com/robert-koch-institut/mex-editor/pull/269/commits/6484cb835c87326b7720cdce123b7d9bf50bed7a

cutoffthetop (approved):

PR #313, opened 2025-04-29
feature/mx-1769-create-new-items-in-UI

ZehraVictoria:
### Added
- create new items in mex-editor UI


cutoffthetop (dismissed):
nice ✨ i see a few spots that could still be de-duplicated between the edit and create pages, but let's merge this as a first iteration.

PR #323, opened 2025-05-14
feature/mx1652 edit identifier fields

rababerladuseladim:
### Added

- allow editing for identifier fields


cutoffthetop (changes requested):
- `InputConfig.pattern  # regular expression for validating text` is ambiguous because it is not only used to validate `editable_text` but also `editable_identifier`. plus, we have cases where 2 `editable_` fields are editable **and** have different patterns (eg URL+Text for links).
- raising a `ValueError` is not very user friendly (looks intimidating, disappears automatically) and they pollute the server logs with something that is not entirely unexpected.
  - especially while typing out an email, i get a bunch of these Error notifications, even though i am not finished typing and haven't clicked submit yet.
- i would actually omit this validation layer entirely. i know the ACs said to validate the identifier pattern, but that was written before we implemented a [summary of the validation errors](https://github.com/robert-koch-institut/mex-editor/blob/0.15.0/mex/editor/edit/state.py#L105-L106) that shows up when you try to submit the changes to the backend. so in my view, that AC is already ticked off.
- if you would rather keep it than scrap it, maybe you can exchange the `ValidationError` for something more friendly, eg a little green checkmark or red cross next to the input field with the tooltip of the matched/missed pattern.

rababerladuseladim:
I removed the pattern validation entirely.

cutoffthetop (approved):

PR #332, opened 2025-06-03
Feature/mx -1726 Highlight required fields in the mex-editor edit view

vyvytranngoc:
### PR Context
In the edit view of the mex-editor, we want to highlight required fields.
required fields can be either single value non-nullable fields or array fields with a minimum length of one

### Added
required input fields are highlighted with a red asterisk


mex/editor/edit/transform.py
```diff
@@ -1,7 +1,7 @@
-from functools import lru_cache
+from functools import lru_cache  # noqa: I001
```
rababerladuseladim: Why no sort?
```suggestion
from functools import lru_cache
```

mex/editor/edit/transform.py
```diff
@@ -227,10 +232,18 @@ def transform_models_to_fields(
             for f in MERGEABLE_FIELDS_BY_CLASS_NAME[e.entityType]
         }
     )
+
+    required_fields = get_required_field_names(type(extracted_items[0]))
```
rababerladuseladim: Are we certain that all `extracted_items` have the same type? If not, we need to get required field for each type.

mex/editor/edit/transform.py
```diff
@@ -247,6 +260,18 @@ def transform_models_to_fields(
     return list(fields_by_name.values())
 
 
+@lru_cache(maxsize=4048)
+def get_required_field_names(model: type[BaseModel]) -> list[str]:
+    """Returns list of required fields."""
```
rababerladuseladim: Please expand the docstring to include arguments

requirements.txt
rababerladuseladim, line 1: Please revert, reflex should not be part of the requirements.txt

rababerladuseladim (changes requested):
Nice work!

My remarks:
- [x] To make the tests in github run, you could set the mex-common dependency to your branch: `pdm add mex-common@git+https://github.com/robert-koch-institut/mex-common.git@feature/mx-1726-highlight-required-fields-in-mex-editor-edit-view`.
- [x] please add a test that checks for the presence of the asterisks
- [x] please add a test making sure that the required fields are what we expect (either here or - if you go for my suggestion in mex-common - in mex-common)
- [x] make the tests pass
- [x] I don't see an asterisk anywhere. Am I overlooking something? image

mex/editor/edit/transform.py
rababerladuseladim: One final idea: you can simply check the fields from `MERGEABLE_FIELDS_BY_CLASS_NAME` for `.is_required()`. That way you don't need to do anything in mex-common.

tests/edit/test_main.py
rababerladuseladim, line 1: neat!

mex/editor/rules/transform.py
```diff
@@ -227,10 +228,21 @@ def transform_models_to_fields(
             for f in MERGEABLE_FIELDS_BY_CLASS_NAME[e.entityType]
         }
     )
+
+    required_fields = (
+        get_required_field_names(extracted_items[0])
+        if extracted_items
+        else additive.entityType
+    )
```
rababerladuseladim: Neat! However, I noticed that the required fields should not change depending on whether we have extracted or additive items. On top, this will never pass the additive item to `get_required_field_names`. I suggest to rename `get_required_field_names` to `get_required_mergeable_field_names` (see suggestion below) and always supply the additive item.

```suggestion
    required_fields = get_required_mergeable_field_names(additive)
```

mex/editor/rules/transform.py
```diff
@@ -247,6 +259,20 @@ def transform_models_to_fields(
...
+
+    Args:
+            model: Model to inspect
+
+    Returns:
+            A list of required fields from given model which are required
+    """
+    required_fields = set(REQUIRED_FIELDS_BY_CLASS_NAME.get(model.entityType, []))
+    mergeable_fields = set(MERGEABLE_FIELDS_BY_CLASS_NAME.get(model.entityType, []))
+    return sorted(required_fields & mergeable_fields)
```
rababerladuseladim: I noticed that we want to get the same required fields, no matter if we provide an additive or an extracted model as input. Otherwise, items that consist only of rules would not have any required fields. Required fields are however defined as the fields required to create a valid merged item. So we should always check the required fields of the merged classes.

(Import `ensure_prefix` from mex.common.transform.)

```suggestion
def get_required_mergeable_field_names(model: AnyExtractedModel | AnyAdditiveModel) -> list[str]:
    """Returns list of required mergeable fields.

    Args:
            model: Model to inspect

    Returns:
            A list of required mergeable fields from given model
    """
    merged_type = ensure_prefix(model.stemType, "Merged")
    required_fields = REQUIRED_FIELDS_BY_CLASS_NAME[merged_type]
    mergeable_fields = MERGEABLE_FIELDS_BY_CLASS_NAME[merged_type]
    return sorted(required_fields & mergeable_fields)
```

requirements.txt
rababerladuseladim: This is still present. Please revert.

CHANGELOG.md
```diff
@@ -8,8 +8,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- Test for MEx-Editor Rules transform
+- fields.py as look-up for required fields for all models
 
 ### Changes
+- required field names for edit mode from 'REQUIRED_FIELDS_BY_CLASS_NAME' in fields.py
+- test_trasnform.py refactored because EditorField has now attribute 'is_required'
+- consent preview uses direct field isIndicatedAtTime as title
```
rababerladuseladim: Changelog of mex-editor is mostly for users, they don't care about any of our internals.

```suggestion
- highlight required fields in edit view with a red asterisk

### Changes
```

rababerladuseladim (changes requested):
Nice and almost there. I noticed that the logic for computing the required fields is not perfect, yet. However, the required changes are minimal.

vyvytranngoc:
Added your requested changes and adapted the test to the changes

mex/editor/rules/transform.py
```diff
@@ -259,17 +255,20 @@ def transform_models_to_fields(
...
 
     Returns:
-            A list of required fields from given model which are required
+            A list of required mergeable fields from given model
     """
-    required_fields = set(REQUIRED_FIELDS_BY_CLASS_NAME.get(model.entityType, []))
-    mergeable_fields = set(MERGEABLE_FIELDS_BY_CLASS_NAME.get(model.entityType, []))
+    merged_type = ensure_prefix(model.stemType, "Merged")
+    required_fields = set(REQUIRED_FIELDS_BY_CLASS_NAME.get(merged_type, []))
+    mergeable_fields = set(MERGEABLE_FIELDS_BY_CLASS_NAME.get(merged_type, []))
```
rababerladuseladim: Why `get` with default `[]`? Don't we want to know if the requested model is not present in the dicts?

tests/conftest.py
```diff
@@ -250,7 +250,6 @@ def load_artificial_extracted_items(
 @pytest.fixture
 def required_fields_extracted_activity() -> list[str]:
     return [
-        "field-title-name",
```
rababerladuseladim: Why is this not required anymore?

requirements.txt
rababerladuseladim: Still not reverted

rababerladuseladim (changes requested):
Almost there.

- [ ] please fix tests

What do you think about looking at this together in order to avoid another review round?

vyvytranngoc:
- dynamic test for checking asterisk combining checks for required and optional merged fields
- complete and passed test runs
- merged main without conflicts

rababerladuseladim (approved):

PR #334, opened 2025-06-06
feature/mx-1880-render-values-in-edit-view

rababerladuseladim:
### Added

* edit view: button for switching between rendered and input view for additive field values

image
image


cutoffthetop (approved):
beautiful ✨

PR #360, opened 2025-07-08
feature/mx-1844 fix temporal options

cutoffthetop:
### Fixed

- only show allowed precisions in temporal field drop-down


esinsj (commented):
The code looks good and works for activities. But for Consent there is still precision of hours, minutes, ...
All time precisions should be adapted, or?

esinsj (approved):
questions answered in office

PR #362, opened 2025-07-08
feature/mx-1866-build-frontend

cutoffthetop:
### Added

- build a static zip file of the frontend on each release

### DIFF
--- a/mex/editor/main.py
+++ b/mex/editor/main.py
@@ -9,17 +9,19 @@ from fastapi import FastAPI, Request, Response
 from fastapi.staticfiles import StaticFiles
 from reflex import constants
 from reflex.config import environment, get_config
-from reflex.reflex import run
+from reflex.reflex import _init, run
 from reflex.state import reset_disk_state_manager
+from reflex.utils.build import setup_frontend_prod
 from reflex.utils.console import set_log_level
 from reflex.utils.exec import get_app_module
 from reflex.utils.export import export
+from reflex.utils.prerequisites import get_compiled_app
 
 from mex.editor.logging import UVICORN_LOGGING_CONFIG, logger
 from mex.editor.settings import EditorSettings
 
 API_DEPLOY_URL_PLACEHOLDER = "https://editor-api"
-WS_DEPLOY_URL_PLACEHOLDER = "ws://editor-api"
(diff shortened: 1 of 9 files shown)


mex/editor/main.py
```diff
@@ -59,20 +66,104 @@ def editor_frontend() -> None:  # pragma: no cover
...
+        "img-src 'self' data:; "
+        "font-src 'self'; "
+        f"connect-src 'self' {websocket_server}"
     )
+    headers["Cache-Control"] = "public, max-age=600"
+    return headers
+
+
+@asynccontextmanager
+async def frontend_lifespan(_app: FastAPI) -> AsyncGenerator[None, None]:
```
rababerladuseladim, line 103: wtf....

rababerladuseladim (changes requested):
I am impressed that you made this works. I bet the reflex people would not be happy if they saw what you did but hey, it works.
- boi, this is hacky and complex. Considering that this requires fiddling with deep reflex internals as well as uses private functions, I really do not want to maintain this. What do you think about switching targets from "compile relfex" to "contain all reflex dependencies in docker image"? `reflex init` during docker build should fetch all the dependencies and already reduce the deploy time down to minutes, I think.
- from my understanding, the reflex dockerfiles already cover our use cases, so we do not need to be hacky: https://github.com/reflex-dev/reflex/blob/main/docker-example/simple-two-port/Dockerfile
	- What's your take on this?
- Test of `test_create_frontend_app_static_serving` fails with

```
E           AssertionError: assert 'application/javascript' == 'text/javascr...charset=utf-8'
E
E             - text/javascript; charset=utf-8
E             + application/javascript

tests\test_main.py:86: AssertionError
```

- 
- docker-compose:
	- `WARN[0000] compose.yaml: the attribute 'version' is obsolete, it will be ignored, please remove it to avoid potential confusion`
	- error during build: `mex-editor-api-1       | WARNING:  ASGI app factory detected. Using it, but please consider setting the --factory flag explicitly.`
		- same for frontend
	- redis and neo4j keep running after stopping docker-compose. Please fix.
 	- any chance to use docker compose in github action tests? That way we'd only define the mex-backend dep in a single location.

PR #368, opened 2025-07-11
feature mx-1888 adds infobox to the ingest page

eckelmannf:
### PR Context

Added a type for the different aux values. Maybe its a bit overengineered but now u cant say sth like 'set_current_aux_provider("best_aux_ever")'. Would be lovely if we can achieve that the string "ldap", "orcid", "wikidata" are only written once (DRY). Only found a solution with an enum, which i didn't like.

### Added

- infobox to ingest page for "ldap" and "wikidata"
- new type for aux_provider instead of str

### Changes

- type of aux provider from `str` to `AuxProvider`

### DIFF
--- a/mex/editor/ingest/main.py
+++ b/mex/editor/ingest/main.py
@@ -3,7 +3,11 @@ from typing import cast
 import reflex as rx
 
 from mex.editor.components import render_value
-from mex.editor.ingest.models import IngestResult
+from mex.editor.ingest.models import (
+    AUX_PROVIDER_LDAP,
+    AUX_PROVIDER_WIKIDATA,
+    IngestResult,
+)
 from mex.editor.ingest.state import IngestState
 from mex.editor.layout import page
 
@@ -137,7 +141,18 @@ def search_input() -> rx.Component:
                     autofocus=True,
                     max_length=100,
                     name="query_string",
-                    placeholder="Search here...",
+                    placeholder=rx.match(
+                        IngestState.current_aux_provider,
+                        (
+                            AUX_PROVIDER_LDAP,
+                            'Please use "*" as placeholder e.g. "Muster*".',
+                        ),
+                        (
+                            AUX_PROVIDER_WIKIDATA,
+                            'Please paste "Concept URI" e.g. "http://www.wikidata.org/entity/Q918501"',
+                        ),
+                        "Search here...",
+                    ),
                     style={
                         "--text-field-selection-color": "",
                         "--text-field-focus-color": "transparent",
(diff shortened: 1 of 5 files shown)


mex/editor/ingest/main.py
```diff
@@ -210,6 +225,25 @@ def search_results() -> rx.Component:
...
 
+def search_infobox() -> rx.Component:
+    """Render some informations about the specific search provider query format."""
+    return rx.match(
+        IngestState.current_aux_provider,
+        (
+            AUX_PROVIDER_LDAP,
+            rx.callout(
+                'Search users by their fullname. Please use "*" as placeholder e.g. "Muster*".'  # noqa: E501
+            ),
```
cutoffthetop: you could use implicit string concatenation here, to avoid the line-too-long error:

```suggestion
            rx.callout(
                "Search users by their fullname. "
                'Please use "*" as placeholder e.g. "Muster*".'
            ),
```

mex/editor/ingest/models.py
```diff
@@ -1,7 +1,21 @@
...
+AUX_PROVIDER_LDAP: Final = "ldap"
+AUX_PROVIDER_ORDIC: Final = "orcid"
+AUX_PROVIDER_WIKIDATA: Final = "wikidata"
+
+AUX_PROVIDERS: list[AuxProvider] = [
+    AUX_PROVIDER_LDAP,
+    AUX_PROVIDER_ORDIC,
+    AUX_PROVIDER_WIKIDATA,
+]
+
```
cutoffthetop, line 21: i know you wrote you didn't like enums, but i also think that's the only way of avoiding to write the strings more than once

```python
class AuxProvider(Enum):
    """Allowed auxiliary providers."""
    LDAP = "ldap"
    ORCID = "orcid"
    WIKIDATA = "wikidata"
```

mex/editor/ingest/main.py
```diff
@@ -137,7 +141,18 @@ def search_input() -> rx.Component:
                     autofocus=True,
                     max_length=100,
                     name="query_string",
-                    placeholder="Search here...",
```
cutoffthetop: these new placeholders don't really fit in the width of the search box, could you either revert the change, shorten the placeholder text or widen the input element?

Screenshot 2025-07-18 at 13-23-29 MEx Editor Ingest

if you decide to keep them, you would need to update the failing test. but maybe it's more stable to match the input field by testid rather than by placeholder anyway.

so instead of `page.get_by_placeholder("Search here...")` you could use `page.get_by_test_id("aux-search-input")` and then add the testid to the input like this `rx.input(..., custom_attrs={"data-testid": "aux-search-input"})`

mex/editor/ingest/main.py
```diff
@@ -210,6 +225,25 @@ def search_results() -> rx.Component:
...
+                'Search users by their fullname. Please use "*" as placeholder e.g. "Muster*".'  # noqa: E501
+            ),
+        ),
+        (
+            AUX_PROVIDER_WIKIDATA,
+            rx.callout(
+                'Search Wikidata by "Concept URI". Please paste URI e.g. "http://www.wikidata.org/entity/Q918501".'
+            ),
+        ),
+    )
```
cutoffthetop, line 230: could you add a test in `tests/ingest/test_main.py` please to verify callouts (and placeholders) show the correct text for the selected provider? (at least for one or two of the providers)

cutoffthetop (changes requested):
nice, congrats to your first mex pull request!

mex/editor/ingest/main.py
eckelmannf: I removed the dynamic placeholder stuff due to its not part of the ticket and i might have done a bit too much.

cutoffthetop (approved):
thanks, ready to merge 🚢

PR #377, opened 2025-07-23
feature/prep reflex update

cutoffthetop:
### Changes

- prep reflex update:
  - avoid byte operators on Vars, use rx.cond instead
  - avoid Var.to_string(), use f-strings to render vars instead
  - avoid magical setters/getters on State, implement them explicitly
  - avoid calling event handlers from other event handlers, use private methods / direct assignments instead
  - avoid kwargs on event handlers, use positional Mapping argument instead
  - avoid getting granian warnings, set env var explicitly
  - avoid defaults for state model fields, making a swap to dataclasses easier
- fix some typing
- de-duplicate pagination component
- de-duplicate ingestion roundtrip tests

### DIFF
--- a/tests/ingest/test_main.py
+++ b/tests/ingest/test_main.py
@@ -28,7 +28,7 @@ def test_aux_tab_section(ingest_page: Page) -> None:
         (
             "ldap",
             "Person",
-            "doesn't exist gs871s9j91k*",
+            "UNKOWNBLABLABLA",
             "L*",
         ),
         (
@@ -40,7 +40,7 @@ def test_aux_tab_section(ingest_page: Page) -> None:
         (
             "orcid",
             "Person",
-            "doesn't exist gs871s9j91k*",
+            "UNKOWNBLABLABLA",
             "Lars",
         ),
     ],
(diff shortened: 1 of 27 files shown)


tests/ingest/test_main.py
```diff
@@ -23,150 +23,124 @@ def test_aux_tab_section(ingest_page: Page) -> None:
...
-    persons_before = result.total
+    result = connector.fetch_extracted_items(
+        None, None, [f"Extracted{stem_type}"], 0, 1
+    )
+    items_before = result.total
 
-    page = ingest_page
-    ldap_tab = page.get_by_role("tab", name="LDAP")
+    # go to the correct tab
+    ldap_tab = page.get_by_role("tab", name=aux_provider)
```
rababerladuseladim:
```suggestion
    aux_provider_tab = page.get_by_role("tab", name=aux_provider)
```

tests/ingest/test_main.py
```diff
@@ -23,150 +23,124 @@ def test_aux_tab_section(ingest_page: Page) -> None:
...
-    pagination_next = page.get_by_test_id("pagination-next-button")
-    pagination_page_select = page.get_by_test_id("pagination-page-select")
-    expect(pagination_previous).to_be_disabled()
-    expect(pagination_next).to_be_disabled()
-    assert pagination_page_select.inner_text() == ""
+    # count the items afterwards
+    result = connector.fetch_extracted_items(
+        None, None, [f"Extracted{stem_type}"], 0, 1
+    )
+    assert result.total >= items_before
```
rababerladuseladim: Hmm, this will pass if we do not add anything. Is there no way to make sure we added an item that was not present before?

rababerladuseladim (changes requested):
Neat, almost ready.

tests/ingest/test_main.py
cutoffthetop: c1fba67

tests/ingest/test_main.py
cutoffthetop: true, fixed in c1fba67

rababerladuseladim (approved):

PR #391, opened 2025-08-06
feature/mx-1889 update to reflex 0.7.3

cutoffthetop:
### PR Context

- this rollsback mx-1978, because of https://github.com/reflex-dev/reflex/pull/4953, in favor of mx-2040
- the issue of the broken node/bun paths has been outsourced to mx-1995

### Changes
- update reflex to 0.7.3
  - change integration of fastAPI endpoints
  - fix some component styling and python typing


eckelmannf (commented):
Unable to run the provided solution. When calling `pdm run editor` the following error occures: "Reflex requires node version 18.18.0 or higher to run, but the detected version is None".
When adding the installed (by reflex init/run) node to the $PATH it works. This should work by default without specifying the $PATH on every machine. Maybe a solution could be to add `environment.PATH.set(environment.PATH + ';%USER%/reflex/fnm/node...')` in https://github.com/robert-koch-institut/mex-editor/blob/main/mex/editor/main.py#L57.

mex/editor/edit/state.py
```diff
@@ -16,17 +17,21 @@ class EditState(RuleState):
     @rx.event
     def load_item_title(self) -> None:
         """Set the item title based on field values."""
-        if self.stem_type and self.fields:
-            self.item_title = transform_fields_to_title(self.stem_type, self.fields)
+        if self.item_id:
```
eckelmannf: Looks like the logic changed here. Should this be part of this PR? Feels like its not part of the reflex update.

mex/editor/merge/state.py
```diff
@@ -130,7 +128,7 @@ def _refresh_merged(self) -> Generator[EventSpec | None, None, None]:
             response = connector.fetch_preview_items(
                 query_string=self.query_strings["merged"],
                 entity_type=entity_type,
-                limit=self.limit,
+                limit=50,
```
eckelmannf: I'm not sure if this is function specific or we should add a constant like DEFAULT_FETCH_LIMIT since every functions uses 50 as value.

mex/editor/edit/state.py
cutoffthetop: absolutely right. i split this off into a new PR: https://github.com/robert-koch-institut/mex-editor/pull/409

mex/editor/merge/state.py
cutoffthetop: fixed in d1fed7e

assets/page-leave-warn-unsaved-changes.js
```diff
@@ -2,11 +2,6 @@ window.mexEditor = {
     changes: false
 }
```
eckelmannf: I added a ticket (MX-2044) for this gone feature.

mex/editor/utils.py
```diff
@@ -26,3 +28,9 @@ async def resolve_editor_value(editor_value: EditorValue) -> None:
     else:
         msg = f"Cannot resolve editor value: {editor_value}"
         raise MExError(msg)
+
+
+def load_settings() -> EditorSettings:
+    """Reset the settings store and fetch the editor settings."""
+    SETTINGS_STORE.reset()
```
eckelmannf, line 35: As far as i understand it the reset method is removing the stored singleton and EditorSettings.get() return a new EditorSetting and stores it. Is `load_settings() == new EditorSettings()` ?

mex/editor/ingest/main.py
```diff
@@ -105,26 +107,21 @@ def ingest_result(result: IngestResult, index: int) -> rx.Component:
     return rx.card(
         rx.vstack(
             rx.hstack(
-                icon_by_stem_type(
-                    result.stem_type,
-                    size=28,
-                    margin="auto 0",
-                    width="28px",
-                ),
+                icon_by_stem_type(result.stem_type, size=24),
```
eckelmannf: Just to make sure that change was intendet: There are missing `margin` and `width` compared to style before. If u wanna keep `margin` and `width`:
```suggestion
                icon_by_stem_type(result.stem_type, size=24, rx.Style(margin="auto 0",width="28px"))
```

mex/editor/ingest/state.py
```diff
@@ -26,23 +25,22 @@ class IngestState(State):
...
+    current_page: int = 1
     current_aux_provider: AuxProvider = AuxProvider.LDAP
     aux_providers: list[AuxProvider] = ALL_AUX_PROVIDERS
     is_loading: bool = True
 
     @rx.var(cache=False)
     def page_selection(self) -> list[str]:
         """Return a list of total pages based on the number of results."""
-        return [f"{i + 1}" for i in range(math.ceil(self.total / self.limit))]
+        return [f"{i + 1}" for i in range(math.ceil(self.total / DEFAULT_FETCH_LIMIT))]
```
eckelmannf: Why add a constant instead of keeping the state value. Does this change imply the user is no longer able to change shown items per page?

mex/editor/ingest/state.py
```diff
@@ -98,13 +101,12 @@ def go_to_next_page(self) -> None:
         self.current_page = self.current_page + 1
 
     @rx.event
-    def ingest_result(self, index: int) -> Generator[EventSpec | None, None, None]:
+    def ingest_result(self, index: int) -> Generator[EventSpec, None, None]:
         """Ingest the selected result to MEx backend."""
         connector = BackendApiConnector.get()
+        model = self.get_value(self.results_extracted[index])  # type: ignore[arg-type]
```
eckelmannf, line 103: nice. bye bye mutbale proxy :)

mex/editor/ingest/main.py
```diff
@@ -285,17 +284,18 @@ def index() -> rx.Component:
             tab_list(),
             rx.spacer(),
             tab_content(),
-            default_value=IngestState.current_aux_provider,
+            default_value=f"{IngestState.current_aux_provider}",
             on_change=[
                 IngestState.set_current_aux_provider,
+                IngestState.reset_query_string,
```
eckelmannf, line 269: Why does the logic change here? Is this required for the version change? Was expecting to see the `reset_query_string` inside the `set_current_aux_provider`. Or was this a bug fix?

mex/editor/merge/state.py
```diff
@@ -130,7 +129,7 @@ def _refresh_merged(self) -> Generator[EventSpec | None, None, None]:
             response = connector.fetch_preview_items(
                 query_string=self.query_strings["merged"],
                 entity_type=entity_type,
-                limit=self.limit,
+                limit=DEFAULT_FETCH_LIMIT,
```
eckelmannf: See comment on https://github.com/robert-koch-institut/mex-editor/blob/304fd7f411650b76922435ff82f4a4eec856aa5c/mex/editor/ingest/state.py#L38

mex/editor/search/main.py
```diff
@@ -172,80 +168,78 @@ def reference_field_filter_identifier(
...
-                    custom_attrs={"data-testid": "reference-field-filter-field"},
-                ),
-                rx.button(
-                    rx.icon("x"),
-                    on_click=[
-                        SearchState.set_reference_filter_field(""),
-                        *full_refresh,
-                    ],
-                ),
+    return rx.vstack(
```
eckelmannf, line 188: Thanks for the styling.

mex/editor/search/main.py
```diff
@@ -260,37 +254,30 @@ def reference_filter_tab() -> rx.Component:
...
     """
-    return rx.tabs.root(
-        rx.tabs.list(
-            rx.tabs.trigger(
-                "Dynamisch",
+    return rx.card(
+        rx.tabs.root(
+            rx.tabs.list(
+                rx.tabs.trigger("Dynamic", value="dynamic"),
+                rx.tabs.trigger("PrimarySource", value="had_primary_source"),
```
eckelmannf: Why removing the data-testid here? Isn't this handy to just click the tab?

mex/editor/components.py
```diff
@@ -142,36 +145,34 @@ def pagination(state: type[IngestState | SearchState]) -> rx.Component:
...
+        ("Distribution", rx.icon("container", size=size, style=style)),
+        ("Organization", rx.icon("building", size=size, style=style)),
+        ("OrganizationalUnit", rx.icon("door_open", size=size, style=style)),
+        ("Person", rx.icon("circle_user_round", size=size, style=style)),
+        ("PrimarySource", rx.icon("hard_drive", size=size, style=style)),
+        ("Resource", rx.icon("archive", size=size, style=style)),
+        ("Variable", rx.icon("box", size=size, style=style)),
+        ("VariableGroup", rx.icon("boxes", size=size, style=style)),
+        rx.icon("file_question", size=size, style=style),
     )
```
eckelmannf, line 266: What are the benefits when using constants? I don't understand why not just `False` and `True`.
```suggestion
        loading=rx.cond(
            value.text,
            False,
            True,
        ),
```

eckelmannf, line 266: Before there was a title tag with the stemtype which is no longer present.
```suggestion
        ("AccessPlatform", rx.icon("app_window", size=size, style=style, title=stem_type)),
        ("Activity", rx.icon("circle_gauge", size=size, style=style, title=stem_type)),
        ("BibliographicResource", rx.icon("book_marked", size=size, style=style, title=stem_type)),
        ("Consent", rx.icon("badge_check", size=size, style=style, title=stem_type)),
        ("ContactPoint", rx.icon("inbox", size=size, style=style, title=stem_type)),
        ("Distribution", rx.icon("container", size=size, style=style, title=stem_type)),
        ("Organization", rx.icon("building", size=size, style=style, title=stem_type)),
        ("OrganizationalUnit", rx.icon("door_open", size=size, style=style, title=stem_type)),
        ("Person", rx.icon("circle_user_round", size=size, style=style, title=stem_type)),
        ("PrimarySource", rx.icon("hard_drive", size=size, style=style, title=stem_type)),
        ("Resource", rx.icon("archive", size=size, style=style, title=stem_type)),
        ("Variable", rx.icon("box", size=size, style=style, title=stem_type)),
        ("VariableGroup", rx.icon("boxes", size=size, style=style, title=stem_type)),
        rx.icon("file_question", size=size, style=style, title=stem_type),
```
Maybe add it to every icon or wrap the icon in a div and add it to the div.

eckelmannf, line 266: Same as above. I dont undersand why using constants here. Maybe u have in mind that those False, True 's are the same as the above ones. But i think this is a bit too much :o

mex/editor/layout.py
```diff
@@ -97,62 +95,74 @@ def nav_bar() -> rx.Component:
...
+                            State.navigate(State.navigate_target),  # type: ignore[misc]
                         ],
                     )
                 ),
                 spacing="3",
+                style=rx.Style(marginTop="1rem"),
+                justify="end",
             ),
+            style=rx.Style(width="calc(340px * var(--scaling))"),
         ),
```
eckelmannf, line 198: Why set the width of the dialog in pixel? I dont like the look of it. It looks way too squashed and when resizing the window and responsivness comes into play fixed widths are not working anymore. IMO it even looks better without the width.

```suggestion
            ),
        ),
```

tests/edit/test_main.py
```diff
@@ -76,6 +76,24 @@ def test_edit_page_renders_heading(
...
+                title=[Text(value="Aktivität 1", language=TextLanguage.DE)],
+            ),
+        ),
+    )
+    page.reload()
+    heading = page.get_by_test_id("edit-heading")
+    expect(heading).to_be_visible()
+    page.screenshot(path="tests_edit_test_main-test_edit_page_renders_new_heading.png")
+    assert re.match(r"New title*", heading.inner_text())
+
```
eckelmannf: Looks like a merging mistake. Its the same as above but with `assert re.match...` instead of `expect...to_have_text(re.compile...`.

```suggestion

```

tests/search/test_main.py
```diff
@@ -208,11 +208,11 @@ def test_reference_filter_fields_for_entity_type(
...
-    dyn_tab = page.get_by_role("tab", name="Dynamisch")
+    dyn_tab = page.get_by_role("tab", name="Dynamic")
     dyn_tab.click()
-    assert page.get_by_test_id("reference-field-filter").is_visible()
+    expect(page.get_by_test_id("reference-field-filter")).to_be_visible()
 
-    # select person entityasdas
+    # select person entity
     entity_types = page.get_by_test_id("entity-types")
     entity_types.get_by_text("Person").click()
```
eckelmannf, line 219: Not part of ur PR but i spotted that the last line (237) of the test is missing an `assert` before `select_item.is_visible()` or it needs to be `expect(select_item).to_be_visible()`.

tests/conftest.py
```diff
@@ -27,20 +30,43 @@
...
 @pytest.fixture
 def client() -> TestClient:
     """Return a fastAPI test client initialized with our app."""
-    with TestClient(app.api, raise_server_exceptions=False) as test_client:
+    with TestClient(api, raise_server_exceptions=False) as test_client:
         return test_client
 
 
+@pytest.fixture(scope="session")
+def browser_context_args(
```
eckelmannf, line 41: No Docstring and why do we need those setting now?

tests/conftest.py
```diff
@@ -27,20 +30,43 @@
...
+        **browser_context_args,
+        "viewport": {
+            "width": 1920,
+            "height": 1080,
+        },
+    }
+
+
+@pytest.fixture(scope="session")
+def browser_type_launch_args(
```
eckelmannf, line 55: No Docstring and why do we need those setting now?

pyproject.toml
```diff
@@ -9,30 +9,30 @@ urls = { Repository = "https://github.com/robert-koch-institut/mex-editor" }
...
     "pytest>=8",
-    "ruff>=0.11",
+    "ruff>=0.12",
     "sphinx>=8",
-    "types-ldap3>=2",
-    "types-PyYAML>=6",
-    "types-requests>=2",
+    "types-ldap3",
+    "types-PyYAML",
+    "types-requests",
```
eckelmannf, line 36: Version numbers are gone. Is this intended?

pyproject.toml
```diff
@@ -53,7 +53,15 @@ show_error_codes = true
 plugins = ["pydantic.mypy"]
 
 [[tool.mypy.overrides]]
```
eckelmannf: When i comment theses overrides out `pdm lint` is still working. Do we really need them? Maybe i dont understand them correctly.

eckelmannf (commented):
Very nice i love FastAPI using decorator. Also very kind of u to incude all feature branches that might be effected. Just having a few questions and comments.
Reflex 0.8 we are comming 🚀

mex/editor/components.py
cutoffthetop: that was my attempt to satisfy https://docs.astral.sh/ruff/rules/boolean-positional-value-in-call

but i agree it's a bit much. solved it with keyword args in https://github.com/robert-koch-institut/mex-editor/pull/391/commits/aba95b03e90f1edb3946c65c0fe7669a0cc1ccfc#diff-4f4bf14e9b5110a5d35778e54e0b744950b20ffd956c7810132ad1db2ac2cb23L77

cutoffthetop: done: 7476370

mex/editor/ingest/main.py
cutoffthetop: yeah this was a small bug fix that caused the backend to complain about wrong query string formats when switching from LDAP to wikidata, for example

mex/editor/ingest/state.py
cutoffthetop: the user was actually never able to do that, the state var was just implying it.

nevertheless, i rolled it back because it isn't connected to the version update: 6d09f15

mex/editor/search/main.py
cutoffthetop: they're back: 935e08d

mex/editor/utils.py
cutoffthetop: yep. this allows the settings to be reloaded too when running in dev mode

pyproject.toml
cutoffthetop: no, you're right :) fixed in 2a5821b

tests/edit/test_main.py
cutoffthetop: fixed 1e681a2

mex/editor/layout.py
cutoffthetop: gone ✔️

tests/search/test_main.py
cutoffthetop: fixed 8755f18

PR #398, opened 2025-08-18
feature/mx-1978 Unsaved changes warning when close browser-tab or -window

eckelmannf:
### Added

- Warning when the create or edit browser window/tab is closed or navigated away

### DIFF
--- a/tests/edit/test_main.py
+++ b/tests/edit/test_main.py
@@ -573,7 +573,10 @@ def test_edit_page_warn_tab_close(edit_page: Page) -> None:
 
     # back to edit site
     page.go_back()
-    page.wait_for_url("**/item/**", wait_until="domcontentloaded")
+    page.wait_for_url("**/item/**")
+    page.wait_for_selector(
+        "[data-testid='new-additive-alternativeTitle-00000000000000']"
+    )
     page.screenshot(
         path="tests_edit_test_main-test_edit_page_warn_tab_close-back_on_edit.png"
     )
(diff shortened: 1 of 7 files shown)


tests/edit/test_main.py
```diff
@@ -558,3 +558,64 @@ def test_required_fields_red_asterisk(
...
+    # change a value
+    page.get_by_test_id("new-additive-alternativeTitle-00000000000000").click()
+    page.get_by_test_id("additive-rule-alternativeTitle-0-text").fill(
+        "new alternative title"
+    )
+
+    handle_dialog_called: list[bool] = []
+
+    def _handle_dialog(dialog: Dialog) -> None:
+        # damn i love python
```
cutoffthetop, line 593: 😬

you could pass the reference to the list along with the handle function to the page.on(dialog, ...) if you wanted to avoid this. i'd find that a bit more readable:

```python
def _handle_dialog(was_called: list[bool], dialog: Dialog) -> None:
   ...

from functools import partial

page.on("dialog", partial(_handle_dialog, handle_dialog_called))
```

cutoffthetop (approved):
small hint for the tests, otherwise good to merge! 🚀

tests/edit/test_main.py
eckelmannf: i would leave it as it as if its okay for you. I never used sth like nonlocal before so i would love to keep this alive, because its special :)

cutoffthetop: fine, since it's just in a test, let's have at it 😄

PR #400, opened 2025-08-19
Feature/mx 1959 toggle all switches

eckelmannf:
### Added

- added a switch to toggle all primary sources and values in the edit page


cutoffthetop (approved):
straight forward solution 👍

PR #401, opened 2025-08-20
feature/mx-1958 textareas for fields with long text

eckelmannf:
### Added

- added textareas for fields listed in mex/editor/models.yaml

### DIFF
--- a/mex/editor/rules/main.py
+++ b/mex/editor/rules/main.py
@@ -234,17 +234,34 @@ def additive_rule_input(
         ),
         rx.cond(
             input_config.editable_text,
-            rx.input(
-                placeholder="Text",
-                value=value.text,
-                on_change=RuleState.set_text_value(field_name, index),
-                style={
-                    "margin": "calc(-1 * var(--space-1))",
-                    "width": "100%",
-                },
-                custom_attrs={
-                    "data-testid": f"additive-rule-{field_name}-{index}-text"
-                },
+            rx.cond(
+                input_config.render_textarea,
+                rx.text_area(
+                    placeholder="Text",
+                    value=value.text,
+                    on_change=RuleState.set_text_value(field_name, index),
+                    style={
+                        "margin": "calc(-1 * var(--space-1))",
+                        "width": "100%",
+                    },
+                    custom_attrs={
+                        "data-testid": f"additive-rule-{field_name}-{index}-text"
+                    },
+                    rows="5",
+                    resize="vertical",
+                ),
+                rx.input(
+                    placeholder="Text",
+                    value=value.text,
+                    on_change=RuleState.set_text_value(field_name, index),
+                    style={
+                        "margin": "calc(-1 * var(--space-1))",
+                        "width": "100%",
+                    },
+                    custom_attrs={
+                        "data-testid": f"additive-rule-{field_name}-{index}-text"
+                    },
+                ),
             ),
         ),
         rx.cond(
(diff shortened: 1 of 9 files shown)


mex/editor/rules/transform.py
```diff
@@ -118,13 +119,16 @@ def _transform_model_to_input_config(  # noqa: PLR0911
             allow_additive=editable,
         )
     if field_name in TEXT_FIELDS_BY_CLASS_NAME[entity_type]:
+        stem_type = cast("Any", ALL_MODEL_CLASSES_BY_NAME[entity_type]).stemType
```
cutoffthetop: ah, sorry. i forgot that ALL_MODEL_CLASSES_BY_NAME, also contains the nested models, which don't have a stemType. that's why you have to do this ugly Any cast here.

tldr; could you just add the stem type as a parameter to this function?

```python
# mex/editor/rules/transform.py
def _transform_model_to_editor_primary_sources(...) -> None:
    ...
    input_config = _transform_model_to_input_config(
        field_name,
        model.entityType,
        model.stemType,
        editable=isinstance(model, AnyAdditiveModel),
    )
    ...

# mex/editor/rules/transform.py
def _transform_model_to_input_config(  # noqa: PLR0911
    field_name: str,
    entity_type: str,
    stem_type: str,
    editable: bool,  # noqa: FBT001
) -> InputConfig:
```

mex/editor/rules/main.py
```diff
@@ -234,17 +234,34 @@ def additive_rule_input(
...
+                rx.text_area(
+                    placeholder="Text",
+                    value=value.text,
+                    on_change=RuleState.set_text_value(field_name, index),
+                    style={
+                        "margin": "calc(-1 * var(--space-1))",
+                        "width": "100%",
+                    },
+                    custom_attrs={
+                        "data-testid": f"additive-rule-{field_name}-{index}-text"
```
cutoffthetop, line 248:
```suggestion
                        "data-testid": f"additive-rule-{field_name}-{index}-textarea"
```

would make it easier for the tests to know what is rendered

--- 

could you also add a test, analogous to `test_edit_page_renders_text_input`, for the textarea please?

cutoffthetop (changes requested):
nice, just two small suggestions

mex/editor/rules/main.py
eckelmannf: I dont think this is a good idea. I understood it that way, that tests wanna find the correct control for clicking, filling, etc and the "_text" suffix indicates that its an text value (like link or identifier). The other main downfall is, that the tests then have to know what kind of field is rendered when they wanna fill in text => leads to alot of changes in existing tests and also every test needs to read the model.yaml to generate the _text or _textarea suffix (or u read it and write it static). Therefore i disagree with the suggested change.

mex/editor/rules/transform.py
eckelmannf: Okay. At some point i gave model as arg (even the func name suggests this), but this would lead to many changes in the tests so i rolled it back. Any thoughts on this? Meanwhile i'll implement as suggested.

cutoffthetop: giving the full model as an argument does not work because this function is cached. but either way, i dont think there is a way around updating the tests

mex/editor/rules/main.py
cutoffthetop: alright, fair points

PR #395, opened 2025-08-13
feature/mx-1961 Reference Filtering

eckelmannf:
### Added
- search: new filter for reference field filtering
- search: tab to switch between new reference field filter and the old had primary source filter
- style.scss to assets (i dont use it anymore, but when i remove it the app won't compile :/)

PR #412, opened 2025-08-29
feature/mx-1531 i18n label and description for fields

eckelmannf:
### PR Context

Currently the locale/.mo files are stored in this repo (/locales). I created a fork on mex-model that includes a github action which converts the .po files to .mo files. When this fork is accepted the path of the locales will change and the will be removed from this repo.
When PR https://github.com/robert-koch-institut/mex-editor/pull/395 is accepted, the translation also need to be added to the reference field filter.

### Added

- labels and descriptions in de and en for edit and create page


mex/editor/layout.py
```diff
@@ -77,6 +78,40 @@ def app_logo() -> rx.Component:
     )
 
 
+def locale_image(locale: str) -> rx.Component:
+    """Render a flag as image for the given locale.
+
+    Args:
+        locale: The locale to render a flag image for.
+    """
+    return rx.image(src=f"/locales/{locale}.png", style={"width": "48px"})
```
rababerladuseladim: Finding these files depends on the current working directory, doesn't it? Or does reflex always look for images in the assets directory? If it depends on the cwd: Could you adapt it, e.g. by providing a path relative to the current file?

rababerladuseladim: The dark mode switch is 24px high, could you make this the same height?
```suggestion
    return rx.image(src=f"/locales/{locale}.png", style={"height": "24px"})
```

assets/locales/de-DE.png
rababerladuseladim, line 1: The pictures have different aspect ratios. Could you adjust that? 
Currently, the language selection looks off: 
image

mex/editor/locale_service.py
```diff
@@ -0,0 +1,170 @@
...
+        Returns:
+            str: Current locale of the app.
+        """
+        return self._current_locale
+
+    def field_label(self, stem_type: str, field_name: str, n: int = 1) -> str:
+        """Get the human readable form the given field.
+
+        Args:
+            stem_type: The type the field belongs to.
```
rababerladuseladim:
```suggestion
            stem_type: The entity type the field belongs to.
```

mex/editor/locale_service.py
```diff
@@ -0,0 +1,170 @@
...
+            field_name: The name of the field.
+            n (optional): Number to pass to ngettext to determine if plural form is
+            used. Defaults to 1.
+
+        Returns:
+            str: The human readable name of the field.
+        """
+        if not self._translation:
```
rababerladuseladim: When is this true? `self._translation` is set upon initialization of the service and should therefore always be present in this function, shouldn't it?

mex/editor/locale_service.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+# IDEA: Propably cleaner to split this into LocalService(str -> translation + ensure
+# fallback translation if str is no valid language or doesnt exist) and
+# FieldHelper(label and description for fields)
+class LocaleService:
+    """A service singleton to control the current locale used by the app."""
+
+    _current_locale: str = LOCALES_AVAILABLE[0]
+    _translation: gettext.GNUTranslations | None = None
```
rababerladuseladim: What's the reason for making these class vars? Could they also be instantiated during init? That way, _translation would not need to allow None.

mex/editor/locale_service.py
```diff
@@ -0,0 +1,170 @@
...
+                lambda x: x != msg_id1,
+            ),
+            (
+                lambda: translation.gettext(msg_id1),
+                lambda x: x != msg_id1,
+            ),
+            (
+                lambda: translation.gettext(field_name),
+                lambda x: x != field_name,
+            ),
```
rababerladuseladim: Oh boi, that is a lot of lambdas. I am having trouble understanding all of these. Could you either add comments to each trans_finder, explaining the current case or move them into dedicated functions, which could also check the translations validity (second lambda of each trans_finder).

mex/editor/locale_service.py
```diff
@@ -0,0 +1,170 @@
...
+        trans_finders = [
+            (
+                lambda: translation.pgettext(stem_type, desc_key),
+                lambda x: x != desc_key,
+            ),
+            (
+                lambda: translation.gettext(desc_key),
+                lambda x: x != desc_key,
+            ),
+        ]
```
rababerladuseladim: please adapt according to the chosen solution for `field_label`.

mex/editor/locale_service.py
```diff
@@ -0,0 +1,170 @@
...
+    _translation: gettext.GNUTranslations | None = None
+
+    @classmethod
+    def get(cls) -> Self:
+        """Get singleton instance of the LocaleService.
+
+        Returns:
+            Self: The LocaleService singleton.
+        """
+        return cast("Self", LOCALE_SERVICE_STORE.load(cls))
```
rababerladuseladim, line 60: Fix typing by using a TypeVar: 

```
from typing import TypeVar

T = TypeVar("T", bound="LocaleService")
```

```suggestion
    def get(cls: type[T]) -> T:
        """Get singleton instance of the LocaleService.

        Returns:
            Self: The LocaleService singleton.
        """
        return LOCALE_SERVICE_STORE.load(cls)
```

mex/editor/rules/transform.py
```diff
@@ -233,20 +234,22 @@ def transform_models_to_fields(
     """
     mergeable_fields = sorted(
         {
-            f
+            (f, e.stemType)
```
rababerladuseladim, line 237: stemTyp should be the same for all extracted items and the additive item. Could you either store it in a dedicated Var or rename `mergeable_fields`, e.g. to `mergeable_fields_and_stem_types`?

mex/editor/locale_service.py
```diff
@@ -0,0 +1,170 @@
...
+
+    def get_locale(self) -> str:
+        """Get the current locale of the app.
+
+        Returns:
+            str: Current locale of the app.
+        """
+        return self._current_locale
+
+    def field_label(self, stem_type: str, field_name: str, n: int = 1) -> str:
```
rababerladuseladim:
```suggestion
    def get_field_label(self, stem_type: str, field_name: str, n: int = 1) -> str:
```

mex/editor/locale_service.py
```diff
@@ -0,0 +1,170 @@
...
+        ]
+
+        for finder in trans_finders:
+            trans = finder[0]()
+            if found_trans := finder[1](trans):
+                break
+
+        return trans if found_trans else camelcase_to_title(field_name)
+
+    def field_description(self, stem_type: str, field_name: str) -> str:
```
rababerladuseladim:
```suggestion
    def get_field_description(self, stem_type: str, field_name: str) -> str:
```

mex/editor/state.py
```diff
@@ -86,6 +88,17 @@ def _update_raw_path(
...
+    @rx.event
+    def change_locale(self, locale: str) -> EventSpec:
+        """Change the current locale to the given one and reload the page.
+
+        Args:
+            locale: The locale to change to.
+        """
+        LocaleService.get().set_locale(locale)
+        self.current_locale = locale
+        return rx.call_script("window.location.reload()")
```
rababerladuseladim, line 100: I am uncertain about the implications of this. Do we loose all state if we trigger a reload? If yes: Should we rather do this dynamically? @cutoffthetop What do you think?

rababerladuseladim (changes requested):
Nice work, neat and efficient!

My remarks:

- [x] could you add tests for
  - [x] `camelcase_to_title`
  - [x] `field_label` (particulary for all the trans_finders)
  - [x] `field_description`
- [x] from a user experience point of view, I would expect to get the hover popup description upon hovering over the entire card and not only over the text. Do you agree? If yes: could you adapt that?
- [ ] I fear that we might forget adding translated labels and descriptions to the reference search filter, once that is done. Are you going to add these as part of this ticket? Or do you prefer to  create a dedicated ticket? I have no preference, I merely want to make sure we do it.

mex/editor/layout.py
eckelmannf: No this is the http request from the client to load the image. All from the assets folder will be available under `/<folder_structure>`.

mex/editor/locale_service.py
eckelmannf: Lovely. Copied the Singleton implementation from the BaseSetting (https://github.com/robert-koch-institut/mex-common/blob/0bf43a801b2c31d680e81452eb6c53365d830962/mex/common/settings.py#L77) class. Maybe we should adapt there too?!

Unfortunately ruff complains about this: Use `Self` instead of custom TypeVar `T` RuffPYI019

mex/editor/locale_service.py
eckelmannf: Good point. I adapt

eckelmannf:
Added ur suggested changes. 
I would like to close this pr and open a new ticket for the reference field translation. Maybe we can add our discussion about how and where we convert and store the .mo files and the resulting implementation adjustments also to this ticket. And we call is sth like "cleanup translation".

mex/editor/locale_service.py
```diff
@@ -0,0 +1,188 @@
...
+            (
+                lambda: translation.gettext(msg_id1),
+                lambda x: x != msg_id1,
+            ),
+            # find field WITHOUT stem_type
+            (
+                lambda: translation.gettext(field_name),
+                lambda x: x != field_name,
+            ),
+        ]
```
rababerladuseladim, line 142: Thank you for adding the comments. I am still having trouble to understand this part and hence fear that it will be hard to maintain.

I have two requests for making this more maintainable:

1. An open question for me is why there are so many different cases to look for a translation. Could we improve this with a better translation file? E.g. with a mandatory stem_type as context? And hence: could we reduce the complexity of this function by making the translation file more consistent? If this is the case: Could you create a ticket with the required steps (steps for cleaning up the translation file, remove functions from `get_field_label`/`get_field_description`, ...)
2. The lambda functions are the largest contributor to making this code hard to understand, imo. Hence, I would like to get rid of the lambdas as far as possible and rather make the encapsulated code explicit. To give an example:

```python
            (
                lambda: translation.npgettext(stem_type, msg_id1, msg_id2, n),
                lambda x: x not in (msg_id1, msg_id2),
            ),
```

would become

```python
msg_id_singular = f"{field_name}.singular"
msg_id_plural = f"{field_name}.plural"

contextual_plural_aware_translation = self._translation.npgettext(stem_type, msg_id_singular, msg_id_plural, n)
if contextual_plural_aware_translation not in (msg_id_singular, msg_id_plural):
    return contextual_plural_aware_translation
```

Could you adapt the code accordingly? (same for `get_field_description`)

mex/editor/locale_service.py
```diff
@@ -0,0 +1,188 @@
...
+        Args:
+            stem_type: The type the field belongs to.
+            field_name: The name of the field.
+
+        Returns:
+            str: The description of the field.
+        """
+        if not self._translation:
+            return ""
+
```
rababerladuseladim, line 165:
```suggestion
```

mex/editor/locale_service.py
```diff
@@ -0,0 +1,188 @@
...
+        translation = self._translation
+        desc_key = f"{field_name}.description"
```
rababerladuseladim, line 167:
```suggestion
        msg_id_description = f"{field_name}.description"
```

mex/editor/state.py
rababerladuseladim: Did you get in touch with @cutoffthetop regarding these questions? In case you did: Could you post the answers here?

One specific case I could think of is that a user creates a new item, switches the language without saving and loses all his edits. Can that happen?

rababerladuseladim:
> Added ur suggested changes. I would like to close this pr and open a new ticket for the reference field translation. Maybe we can add our discussion about how and where we convert and store the .mo files and the resulting implementation adjustments also to this ticket. And we call is sth like "cleanup translation".

Sure, could you create a ticket a comment the ticket number in this PR?

mex/editor/locale_service.py
```diff
@@ -0,0 +1,188 @@
...
+    return LOCALES_LABEL_MAPPING[locale]
+
+
+# IDEA: Propably cleaner to split this into LocalService(str -> translation + ensure
+# fallback translation if str is no valid language or doesnt exist) and
+# FieldHelper(label and description for fields)
+class LocaleService:
+    """A service singleton to control the current locale used by the app."""
+
+    _current_locale: str = ""
```
rababerladuseladim, line 50: Hmm, the default is an "illegal" value for the `_current_locale`, isn't it? Illegal in that sense that the `LocaleService` breaks if it is initialized with `""`. What do you think about introducing an Enum for the allowed values? That way, type checking could make sure that the init/set_locale methods are called with the correct arguments and the service would be easier to re-use, since devs don't need to browse the code to find out allowed strings.

tests/test_locale_service.py
```diff
@@ -0,0 +1,156 @@
...
+        ("en-US", "AnyStemType", "alternativeTitle", False, "Alternative title"),
+        ("en-US", "AnyStemType", "alternativeTitle", True, "Alternative titles"),
+        ("de-DE", "UnknownStemType", "unknownField", False, "Unknown Field"),
+        ("de-DE", "UnknownStemType", "unknownField", True, "Unknown Field"),
+        ("en-US", "UnknownStemType", "unknownField", False, "Unknown Field"),
+        ("en-US", "UnknownStemType", "unknownField", True, "Unknown Field"),
+    ],
+    ids=[
+        "de:Resource.conformsTo(singular)",
+        "de:Resource.conformsTo(pluaral)",
```
rababerladuseladim, line 53:
```suggestion
        "de:Resource.conformsTo(plural)",
```

tests/test_locale_service.py
```diff
@@ -0,0 +1,156 @@
...
+
+@pytest.mark.parametrize(
+    ("locale", "stem_type", "field_name", "expected_description"),
+    [
+        (
+            "de-DE",
+            "Activity",
+            "abstract",
+            "Kurze Beschreibung des Kontexts, in dem die Daten verarbeitet wurden",
+        ),
```
rababerladuseladim, line 101: To keep parameters and their ID together and hence make tests easier to adapt (feel free to adapt, I am also fine with the current implementation): 

```suggestion
        pytest.param(
            "de-DE",
            "Activity",
            "abstract",
            "Kurze Beschreibung des Kontexts, in dem die Daten verarbeitet wurden",
             id="de:Activity.abstract"
        ),
```

tests/test_locale_service.py
```diff
@@ -0,0 +1,156 @@
...
+        "de:Activity.abstract",
+        "en:Activity.abstract",
+        "de:Resource.accessPlatform",
+        "en:Resource.accessPlatform",
+        "de:Person.affiliation",
+        "en:Person.affiliation",
+        "de:AnyStemType.alternativeTitle",
+        "en:AnyStemType.alternativeTitle",
+        "de:AnyStemType.identifier",
+        "en:AnyStemType.identifier",
```
rababerladuseladim, line 142: Could you add a test where no description is found?

rababerladuseladim (changes requested):
Nice, thank you for all the tests. I have a few more remarks.

eckelmannf:
followup cleanup translation ticket id is: MX-2021 (.../browse/MX-2021)

mex/editor/state.py
eckelmannf: No I didn't get in touch with him. I expected github to notify him that he is mentioned here. I'll contact him.
Your described behaviour is absolutly correct. If the user changes the languge the current state is lost. Since the AC doesn't say anthing about this i used the simplest solution :)

mex/editor/locale_service.py
eckelmannf: Yeah good point. Was also thinking about this while creating the constants. I'll change this, thx for suggesting.

mex/editor/state.py
cutoffthetop: loosing unsaved changes upon a language-switch is definitely undesired behavior. but since mx-1978 is already merged, shouldn't that prevent the user from reloading if there are unsaved changes? if not, please either fix it in this PR or create a follow-up ticket

mex/editor/locale_service.py
eckelmannf: Since u really hate it i changed it. But i dont see the real benefits here. I asume when u maintain this function u add or remove translation finders, which was kinda more easy with the array of finders. I can see u dont like the 2nd lamda to check if sth was found. that could be one func for sure. w/e i adjusted to ifs :)
I'm not sure if we should change the translation file since as far as i unterstood external partners are using this file too. I already talked to Franziska and changed some entries that looked kinda wrong . Looking for fieldnames that have no context is handy for fields that are trasnlated the same for every entitytype e.g. `alternativeTitle`.

eckelmannf:
Closing this PR due to complete rework in https://github.com/robert-koch-institut/mex-editor/pull/422. This PR will not be merged!

PR #422, opened 2025-09-12
Feature/mx 1531 rework i18n field label description

eckelmannf:
### PR Context
This is a rework of https://github.com/robert-koch-institut/mex-editor/pull/412
main differences are:
- locale service
  - no state for current language (moved to `State` https://github.com/robert-koch-institut/mex-editor/compare/feature/mx-1531-REWORK-i18n-field-label-description?expand=1#diff-6bb38edb48fa77ed98e9bc70528eca2969db564bc8d9764f497ec61bfe0a0653R16)
  - label and description translation now also needs the locale_id (removed the strenum because not supported in reflex state and when loading languages dynamic (see next point), an enum is not suitable)
  - searching for .mo files at startup (https://github.com/robert-koch-institut/mex-editor/compare/feature/mx-1531-REWORK-i18n-field-label-description?expand=1#diff-6dbefbed0e764e0753bd6f5a2319fa4728eeacea8692bc1a2864f69f7d18f59aR71) and provide found locales (https://github.com/robert-koch-institut/mex-editor/compare/feature/mx-1531-REWORK-i18n-field-label-description?expand=1#diff-6dbefbed0e764e0753bd6f5a2319fa4728eeacea8692bc1a2864f69f7d18f59aR80)
  - lazy loading of translations when needed (https://github.com/robert-koch-institut/mex-editor/compare/feature/mx-1531-REWORK-i18n-field-label-description?expand=1#diff-6bb38edb48fa77ed98e9bc70528eca2969db564bc8d9764f497ec61bfe0a0653R16)
- RuleState
  - computed var for translated fields based on fields, stem_type and current_locale (https://github.com/robert-koch-institut/mex-editor/compare/feature/mx-1531-REWORK-i18n-field-label-description?expand=1#diff-ee850869939201565c13d6e48bae9fec364124adbd1bcf3d2cb84edf0340f1e2R50)

- TODO: When merging main, ensure to add reference field filter translation!

### Added
- labels and descriptions in de and en for edit and create page


eckelmannf:
I fear we have to add another select component to adress the translation of fields in the reference filter. The rx.select only allows one str (fieldname) but no way to give a label with it. This library/component has options: dict[str, str]. https://github.com/woernerm/dynoselect/tree/main

mex/editor/rules/state.py
```diff
@@ -43,6 +52,30 @@ class RuleState(State):
...
+                FieldTranslation(
+                    field=field,
+                    label=locale_service.get_field_label(
+                        self.current_locale, self.stem_type, field.name
+                    ),
+                    description=locale_service.get_field_description(
+                        self.current_locale, self.stem_type, field.name
+                    ),
+                )
+                for field in fields
```
rababerladuseladim, line 75: Why not like this?

```suggestion
                for field in self.fields
```

mex/editor/rules/transform.py
```diff
@@ -231,22 +231,21 @@ def transform_models_to_fields(
...
-
     required_fields = get_required_mergeable_field_names(additive)
     fields_by_name = {
         field_name: EditorField(
             name=field_name,
             primary_sources=[],
             is_required=field_name in required_fields,
         )
-        for field_name in mergeable_fields
+        for field_name, stem_type in mergeable_fields_with_stemtype
```
rababerladuseladim: stem_type is not needed anymore, could your revert these changes?

mex/editor/locale_service.py
```diff
@@ -0,0 +1,177 @@
...
+from mex.common.context import SingleSingletonStore
+
+
+class MexLocale(TypedDict):
+    """Represents a locale with id and label."""
+
+    id: str
+    label: str
+    filepath: str
+
```
rababerladuseladim: This could also be a pydantic model and hence allow access by attributes names (benefit e.g. easier to refactor, more in sync with our codebase) instead of a TypedDict. To do so, import BaseModel `from mex.common.models import BaseModel`.

mex/editor/locale_service.py
```diff
@@ -0,0 +1,177 @@
...
+from mex.common.context import SingleSingletonStore
+
+
+class MexLocale(TypedDict):
+    """Represents a locale with id and label."""
+
+    id: str
+    label: str
+    filepath: str
```
rababerladuseladim: Couldn't this be a `Path`? I'd expect handling it would be easier, as no conversion to str is required and `open()` is directly supported.

mex/editor/locale_service.py
```diff
@@ -0,0 +1,177 @@
...
+        msg_id1 = f"{field_name}.singular"
+        msg_id2 = f"{field_name}.plural"
+
+        # find plural or singular field for stem_type (most specific)
+        if (
+            translated_fieldname := translation.npgettext(
+                stem_type, msg_id1, msg_id2, n
+            )
+        ) not in (msg_id1, msg_id2):
+            return translated_fieldname
```
rababerladuseladim: Oh boi, half of what I wanted to say was apparently lost 🙈 

My specific ideas were: 

1. rename msg_id1 to msg_id_singular and msg_id2 to msg_id_plural to make obvious what these variables contain.
2. rename the returned variable of each if statement to a specific var name for the current case, e.g. `stem_type_specific_plural_aware_translation` for the current if block, and consecutively get rid of the comment as that is now contained in the var name.

I really don't want to bug you any further with this as I see myself as to blame for not getting my point across. Feel free to adapt if you want to, I won't insist on it.

mex/editor/locale_service.py
```diff
@@ -0,0 +1,177 @@
...
+
+    _available_locales: dict[str, MexLocale] = {}
+    _translations: dict[str, gettext.GNUTranslations] = {}
+
+    def __init__(self) -> None:
+        """Initalize with all found locales in `LOCALE_FOLDER_PATH`.
+
+        Args:
+            self (Self): Self
+        """
```
rababerladuseladim:
```suggestion
        """Initalize with all found locales in `LOCALE_FOLDER_PATH`."""
```

mex/editor/locale_service.py
```diff
@@ -0,0 +1,177 @@
...
+        files = list(LOCALE_FOLDER_PATH.glob("*.mo"))
+        self._available_locales = {
+            mo_file.stem: MexLocale(
+                id=mo_file.stem,
+                label=LOCALES_LABEL_MAPPING[mo_file.stem],
```
rababerladuseladim: The number of available languages is restricted by the LOCALES_LABEL_MAPPING, isn't it? If there would be a new file, e.g. for french, this code would fail because the lookup in `LOCALES_LABEL_MAPPING` fails, doesn't it? 

If that is true: I would suggest to adapt the code and iterate over all keys of LOCALES_LABEL_MAPPING, check for the presence of the required mo file and finally build the MexLocale.

rababerladuseladim (changes requested):
Very nice and very reflexy!

One remark regarding the UI: When hovering over the language selector, the mouse cursor becomes a text cursor, while I'd expect a similar UI as for the user button, which is, a regular mouse arrow and a change of highlighting upon hover. Could you adapt?

mex/editor/rules/state.py
eckelmannf: Due to error. get_value ensures MockedProxies are unwrapped correctly.

mex/editor/rules/transform.py
rababerladuseladim: please also revert the variable name `mergeable_fields_with_stemtype` to `mergeable_fields`. The state of the file should be the same as on main.

mex/editor/locale_service.py
rababerladuseladim: This is not adressed, yet. The docstring implies that I can add a new language to the `LOCALE_FOLDER_PATH` and this code will discover and load it. In reality, this code will crash as long as the file is not also represented in `LOCALES_LABEL_MAPPING`. One solution would be (as suggested in my previous comment) to iterate over LOCALES_LABEL_MAPPING, build the Path of each expected file and check for it's presence. On top, the docstring needs adapting to describe the new behaviour (initialize for all locales in `LOCALES_LABEL_MAPPING`).

mex/editor/locale_service.py
rababerladuseladim: ❤️

rababerladuseladim (changes requested):
Neat!

What still remains is my remark regarding the UI: When hovering over the language selector, the mouse cursor becomes a text cursor, while I'd expect a similar UI as for the user button, which is, a regular mouse arrow and a change of highlighting upon hover. Could you adapt?

rababerladuseladim (approved):
Nice, thank you. 

The LocaleService initialization was still based on the found files and not on the hard coded mapping. I took the liberty of adapting that.

mex/editor/locale_service.py
rababerladuseladim: implemented in 69ad2e5

PR #425, opened 2025-09-15
feature/mx-2022 save indicator in edit and create page

eckelmannf:
### Added
- submit button is disabled and contains a spinner while submitting


cutoffthetop (approved):
👌

PR #426, opened 2025-09-15
hotfix/mx-2042 incorrect default language for link and text

eckelmannf:
### PR Context
Found a bug while editing MX-1990. Simply changed the defualt badge value to none which was used by default even when de was set as default.

### Fixed
- badge default for link and text


cutoffthetop (approved):
nice fix

PR #447, opened 2025-10-17
feature/mx-2021 translate reference filter field

eckelmannf:
### PR Context
Not all fields are translated and diffferent field names translate to the same word/phrase. I created a ticket (mx-2070) for this.

### Added
- CustomSelect-Component that has items with value and label instead of simple strings

### Changes
- Field of the ReferenceFilter is now translated

### DIFF
--- a/mex/editor/search/custom_select.py
+++ b/mex/editor/search/custom_select.py
@@ -1,0 +1,98 @@
+from typing import Any
+
+import reflex as rx
+from reflex.components import Component
+from reflex.components.radix.themes.components.select import (
+    HighLevelSelect,
+    SelectContent,
+    SelectGroup,
+    SelectItem,
+    SelectLabel,
+    SelectRoot,
+    SelectTrigger,
+)
+from reflex.vars.base import Var
+
+
+class CustomHighLevelSelect(HighLevelSelect):
+    """High level wrapper for the Select component."""
+
+    @classmethod
+    def create_value_label_select(
+        cls,
+        items: list[dict[str, str]] | Var[list[dict[str, str]]],
+        **props,  # noqa: ANN003
+    ) -> Component:
+        """Create a select component. THIS IS COPY PASTE FROM HighLevelSelect!
+
+        Args:
+            items: The items (as dict with "label" and "value") of the select.
+            **props: Additional properties to apply to the select component.
+
+        Returns:
+            The select component.
+        """
+        trigger_prop_list = [
+            "id",
+            "placeholder",
+            "variant",
+            "radius",
+            "width",
+            "flex_shrink",
+            "custom_attrs",
+        ]
+
+        content_props = {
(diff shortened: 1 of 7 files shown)


mex/editor/search/custom_select.py
```diff
@@ -0,0 +1,98 @@
...
+    SelectGroup,
+    SelectItem,
+    SelectLabel,
+    SelectRoot,
+    SelectTrigger,
+)
+from reflex.vars.base import Var
+
+
+class CustomHighLevelSelect(HighLevelSelect):
```
cutoffthetop: "Custom" does not tell me what this does differently than "HighLevelSelect". could you choose a name that alludes to the difference and document it in the class or constructor docstring?

mex/editor/search/state.py
```diff
@@ -54,6 +55,9 @@ def _build_had_primary_source_refresh_params(
     }
 
 
+locale_service = LocaleService.get()
```
cutoffthetop: if `LocaleService` is a singleton anyway, and `get` returns this singular instance, why assign it to a module constant? you can just use `LocaleService.get()` where you need it

mex/editor/search/state.py
```diff
@@ -142,11 +146,28 @@ def all_fields_for_entity_types(self) -> list[str]:
...
-        fields_by_type = [
-            REFERENCE_FIELDS_BY_CLASS_NAME[ensure_prefix(entity_type, "Extracted")]
+        fields_with_type = [
+            [
+                {
+                    "value": field,
+                    "label": locale_service.get_field_label(
+                        self.current_locale, entity_type, field
+                    ),
+                }
```
cutoffthetop: could you create a model for this dict - and then indicate in `CustomHighLevelSelect` that it is expecting its menu items to have this model as a type?

tests/conftest.py
```diff
@@ -55,10 +54,10 @@ def browser_context_args(
 def browser_type_launch_args(
     browser_type_launch_args: dict[str, Any],
 ) -> dict[str, Any]:
-    """Run the playwright browser in headed mode locally and in headless mode in CI."""
+    """Run the playwright browser always in headless mode."""
     return {
         **browser_type_launch_args,
-        "headless": os.getenv("CI") is not None,
+        "headless": True,
     }
```
cutoffthetop: let's just remove the launch args then

```suggestion
```

cutoffthetop (changes requested):
nice, i have a few requests/ideas though

tests/conftest.py
eckelmannf: Yeah i was also thinkin the same. But i thought it's kinda handy when u wanna change some settings and don't have to google it again. I removed it, if u think we should keep it, i can bring it back :)

cutoffthetop (approved):
looks good!

PR #467, opened 2025-11-10
Feature/mx 1994 translate UI text

eckelmannf:
### PR Context
A lot of ui elements were searched by strings thats are now localized. I replaced them by data-test-ids or regexs.

### Added
- Decorator for localized labels that creates a `ComputedVar[str]` that depends on `State.current_locale` and calls locale_service.get_text with a `label_id`
- Localized labels for every page, component and layout

### Changes
- Removed .mo files and converting .po files at runtime with polib
- tests:
    - replaced all locators thats searched for localized strings with data_test_id searches
    - replaced localized string compares with regexs


mex/editor/search/state.py
```diff
@@ -56,6 +57,9 @@ def _build_had_primary_source_refresh_params(
     }
 
 
+locale_service = LocaleService.get()
+
+
```
cutoffthetop:
```suggestion
```

mex/editor/label_var.py
```diff
@@ -0,0 +1,109 @@
...
+def label_var(
+    fget: None = None,
+    label_id: str = "",
+    initial_value: Any | types.Unset = types.Unset(),  # noqa: ANN401, B008
+    cache: bool = True,  # noqa: FBT001, FBT002
+    deps: list[str | Var] | None = None,
+    interval: datetime.timedelta | int | None = None,
+    backend: bool | None = None,  # noqa: FBT001
+    **kwargs,  # noqa: ANN003
+) -> Callable[[Callable[[StateT], ReturnT]], ComputedVar[str]]: ...  # pyright: ignore [reportInvalidTypeVarUse]
```
cutoffthetop:
```suggestion
) -> Callable[[Callable[[StateT], ReturnT]], ComputedVar[str]]: ...
```

pyright is not officially part of our stack, so it does not need to be happy about this line 🙃

mex/editor/label_var.py
```diff
@@ -0,0 +1,109 @@
...
+) -> ComputedVar | Callable[[Callable[[StateT], Any]], ComputedVar]:
+    """A decorator to translate the given label_id base on the current locale.
+
+    Args:
+        fget: The getter function.
+        label_id: The msgid to search for in the currents locale .po file
+        initial_value: The initial value of the computed var.
+        cache: Whether to cache the computed value.
+        deps: Explicit var dependencies to track.
+        auto_deps: Whether var dependencies should be auto-determined.
```
cutoffthetop:
```suggestion
```

mex/editor/search/state.py
```diff
@@ -392,6 +396,42 @@ def get_available_primary_sources(self) -> Generator[EventSpec, None, None]:
...
 
+    @label_var(label_id="search.search_input.placeholder")
+    def label_search_input_placeholder(self) -> None:
+        """Label for search_input.placeholder."""
+
+    @label_var(label_id="search.entitytype_filter.title")
+    def label_entitytype_filter_title(self) -> None:
+        """Label for entitytype_filter.title."""
+
+    @label_var(label_id="search.referencefield_filter.field.placholder")
```
cutoffthetop:
```suggestion
    @label_var(label_id="search.reference_field_filter.field.placeholder")
```

mex/editor/locale_service.py
```diff
@@ -51,15 +52,15 @@ def get(cls) -> Self:
 
     def __init__(self) -> None:
         """Initialize with all available locales."""
-        for mo_file in cast("Path", (files("mex.model") / "i18n")).glob("*.mo"):
-            locale = mo_file.name.removesuffix(".mo")
+        for po_file in (Path.cwd() / "locales").glob("*.po"):
```
cutoffthetop: that the editor is launched from the repo root, and that we can use `cwd` is not safe assumption. i would prefer to stick to importlib

mex/editor/locale_service.py
```diff
@@ -68,11 +69,35 @@ def get_available_locales(self) -> Sequence[MExLocale]:
...
         return self._translations[locale_id]
 
+    def get_ui_label(self, locale_id: str, msg_id: str) -> str:
+        """Get the text for a given locale_id and the msg_id.
+
+        Args:
+            locale_id (str): The locale to use.
+            msg_id (str): The message id of the message to get the text for.
+
+        Returns: The message of the msg_id for the given locale_id.
```
cutoffthetop:
```suggestion
        Returns:
            The message of the msg_id for the given locale_id.
```

mex/editor/state.py
```diff
@@ -30,32 +31,47 @@ class State(rx.State):
...
+
+        def _translate(item: NavItem) -> NavItem:
+            return NavItem(
+                title=locale_service.get_ui_label(self.current_locale, item.title),
+                path=item.path,
+                raw_path=item.raw_path,
+                underline=item.underline,
+            )
+
+        return [_translate(item) for item in self.nav_items]
```
cutoffthetop:
```suggestion
        return [
            NavItem(
                title=locale_service.get_ui_label(self.current_locale, item.title),
                **item.dict(exclude={"title"}),
            )
            for item in self.nav_items
        ]
```

tests/create/test_main.py
```diff
@@ -26,7 +26,7 @@ def test_create_page_updates_nav_bar(create_page: Page) -> None:
     page.screenshot(path="tests_create_test_main-test_create_page_updates_nav_bar.png")
     expect(nav_bar).to_be_visible()
     nav_item = nav_bar.locator(".nav-item").all()[1]
-    expect(nav_item).to_contain_text("Create")
+    expect(nav_item).to_contain_text(re.compile(r"\w+"))
```
cutoffthetop: couldn't we use `(Create|Erstellen)` here, if we're not sure which languages is used?

eckelmannf: I tried to remove every "search by ui label" because i assume that the labels are managed by mtdmg and they change it when they think another text is more appropiate. So it would lead to failing tests when the ui text changes or a new language is added. Pros and Cons on every side, i thought labels are not so important and our tests should more focus on funcionality.

mex/editor/state.py
eckelmannf: <3

mex/editor/search/state.py
eckelmannf: If u change label_ids plz make sure to also change the msgid in the po files.

tests/create/test_main.py
cutoffthetop: how about this then? this would at least prevent getting an untranslated msg_id in the navbar

```suggestion
    expect(nav_item).to_contain_text(
        LocaleService.get().get_ui_label("de-DE", "layout.nav_bar.create_navitem")
    )
```

eckelmannf: Mmhhh... I don't like the fact u're asuming current_locale is "de". I liked the (Create|Erstellen) approach more. Maybe we build a helper that gets the regex for a given label_id by collecting the translation from every locale file. Since we use polib we can parse the content of the po files pretty easy and search for specific msgid in each file. What u think about this?

cutoffthetop: but the current_locale is deterministically "de", when would it not be? (except the test explicitily clicks on the drop-down and changes it)

eckelmannf: I wrote the util function for build label regex based on label_id. It felt too bad to hardcode "de" into the function since we load the locales dynamic. I implmeneted the (deLocal|enLocal|xLocal) version 🤠

tests/test_utils.py
```diff
@@ -49,3 +52,14 @@ async def test_resolve_editor_value(
...
+
+def build_pagination_regex(current: int, total: int) -> re.Pattern:
+    return re.compile(rf"\w+\s{current}\s\w+\s{total}\s\w+")
+
+
+def build_ui_label_regex(label_id: str) -> re.Pattern:
+    service = LocaleService.get()
+    return re.compile(
+        f"({'|'.join(re.escape(service.get_ui_label(locale.id, label_id)) for locale in service.get_available_locales())})"
+    )
```
cutoffthetop: these should go into tests/conftest. tests/test_utils is where the tests for mex.editor.utils live

mex/editor/locale_service.py
```diff
@@ -68,11 +72,36 @@ def get_available_locales(self) -> Sequence[MExLocale]:
...
+            error_msg = f"Given 'locale_id' ({locale_id}) is not valid. Valid values "
+            f"are: {', '.join([loc.id for loc in self.get_available_locales()])}"
+            raise ValueError(error_msg)
+
         if locale_id not in self._translations:
-            with self._available_locales[locale_id].filepath.open("rb") as mo_file:
-                self._translations[locale_id] = GNUTranslations(mo_file)
+            filename = f"{locale_id}.po"
+            model_po = polib.pofile(
+                cast("Path", (files("mex.model") / "i18n")) / filename
```
cutoffthetop: could you make this consistent with `_editor_locale_path` please? define either both on the class or both in the function scope

cutoffthetop (changes requested):
nice! just 2 consistency requests. and those flakey tests :/

cutoffthetop (approved):

PR #491, opened 2025-12-02
feature/mx-1995 bun usage

cutoffthetop:
### Changes

- upgrade reflex to version 0.7.14
- fix and/or evade type checking issues with reflex event handlers
- use pytest base_url plugin to make editor url configurable for testing

### Removed

- dropped npm support in favor of bun

### Fixed

- fixed bun cache options for windows


eckelmannf:
Looks very good. Finally we could make bun work 🎆 

I dont unterstand why sometimes cast("Eventhandler", ...)(*args) is used and sometimes #ignore. I dont like the cast call since its pretty unreadable, adds alot of code for nothing (since its passed into reflex so we dont care about typing at this point) and it might add alot of work for us when updating reflex versions. In all their examples the never use types, i assume because its hard to tell if its the state type or object (same for vars and 'normal' values o0).

pyproject.toml
```diff
@@ -10,16 +10,17 @@ requires-python = ">=3.11,<3.12"
 dependencies = [
     "async-lru>=2",
     "babel>=2",
-    "fastapi>=0.117",
+    "fastapi>=0.128",
     "mex-common>=1.5.1,<1.6",
-    "polib==1.2.0",
+    "polib>=1",
```
eckelmannf: Why change this? It's only tested with polib 1.2.0. The polib repo has no changes for over a year. Since there are more and more attacks via github repo maintainer changes i would prefer to pinpoint to a specifc version, rather than sayin "minimum is ...". I also would consider it for all the other packages.

eckelmannf (approved):
Looks good. Tried to get rid of `cond(value, value, fallback)` but couldn't make it work :(

PR #493, opened 2025-12-03
feature/mx-2102 edit create local storage cleanup

eckelmannf:
### Added

- LocalStorage to store drafts and edits local on browser side.

### Changes

- Using href navigation instead of navigation via code (`page.navigate()`).

### Removed

- Dialogs for navigating away while having changes on create or edit page.


mex/editor/create/main.py
```diff
@@ -50,20 +50,73 @@ def create_title() -> rx.Component:
...
 
+def discard_draft_button() -> rx.Component:
+    """Render a button to show discard draft dialog."""
+    return rx.cond(
+        CreateState.has_local_draft,
+        rx.alert_dialog.root(
+            rx.alert_dialog.trigger(
+                rx.button(
+                    CreateState.label_discard_draft_button,
+                    color_scheme="red",
```
cutoffthetop: we use `tomato` for error/cancel and `jade` for success/submit

```suggestion
                    color_scheme="tomato",
```

mex/editor/create/main.py
```diff
@@ -50,20 +50,73 @@ def create_title() -> rx.Component:
...
+                        rx.button(
+                            CreateState.label_discard_draft_dialog_cancel_button,
+                            variant="soft",
+                            color_scheme="gray",
+                        ),
+                    ),
+                    rx.alert_dialog.action(
+                        rx.button(
+                            CreateState.label_discard_draft_dialog_discard_button,
+                            color_scheme="red",
```
cutoffthetop:
```suggestion
                            color_scheme="tomato",
```

mex/editor/layout.py
```diff
@@ -70,19 +73,105 @@ def language_switcher() -> rx.Component:
...
+                    rx.menu.trigger(
+                        rx.badge(
+                            RuleState.draft_summary.count,
+                            style=rx.Style(
+                                {
+                                    "align-self": "center",
+                                    "margin-left": "-1em",
+                                    "cursor": "pointer",
+                                    "border": "1px solid transparent",
+                                }
```
cutoffthetop: rx.Style allows keyword args as well. makes styles a little more compact:

```suggestion
                                alignSelf="center",
                                marginLeft="-1em",
                                cursor="pointer",
                                border="1px solid transparent",
```

mex/editor/edit/main.py
```diff
@@ -36,13 +36,60 @@ def toggle_all_switch() -> rx.Component:
...
 
+def discard_changes_button() -> rx.Component:
+    """Render a button to show discard changes dialog."""
+    return rx.cond(
+        EditState.has_changes,
+        rx.alert_dialog.root(
+            rx.alert_dialog.trigger(
+                rx.button(
+                    EditState.label_discard_changes_button,
+                    color_scheme="red",
```
cutoffthetop:
```suggestion
                    color_scheme="tomato",
```

mex/editor/edit/main.py
```diff
@@ -36,13 +36,60 @@ def toggle_all_switch() -> rx.Component:
...
+                        rx.button(
+                            EditState.label_discard_changes_dialog_cancel_button,
+                            variant="soft",
+                            color_scheme="gray",
+                        ),
+                    ),
+                    rx.alert_dialog.action(
+                        rx.button(
+                            EditState.label_discard_changes_dialog_discard_button,
+                            color_scheme="red",
```
cutoffthetop:
```suggestion
                            color_scheme="tomato",
```

mex/editor/rules/state.py
```diff
@@ -257,82 +367,94 @@ def toggle_primary_source(
...
         enabled: bool,  # noqa: FBT001
-    ) -> None:
+    ) -> Generator[EventSpec, None, None]:
         """Toggle the `enabled` flag of a primary source."""
         for primary_source in self._get_primary_sources_by_field_name(field_name):
             if primary_source.name.href == href:
                 primary_source.enabled = enabled
-                self.set_current_page_has_changes(True)  # type: ignore[misc]
+                yield RuleState.update_local_state
+                # yield RuleState.update_local_edit
```
cutoffthetop: ?

mex/editor/components.py
```diff
@@ -65,12 +81,11 @@ def render_identifier(value: EditorValue) -> rx.Component:
     return rx.skeleton(
         rx.link(
             value.text,
-            on_click=State.navigate(value.href),  # type: ignore[misc]
+            href=value.href,
```
cutoffthetop: 😻

mex/editor/layout.py
```diff
@@ -70,19 +73,105 @@ def language_switcher() -> rx.Component:
...
+                        rx.foreach(
+                            RuleState.draft_summary.drafts, render_draft_menu_item
+                        )
+                    ),
+                ),
+            ),
+            link,
+        ),
+        link,
+    )
```
cutoffthetop: merge hiccup?

```suggestion
```

mex/editor/rules/state.py
```diff
@@ -39,22 +43,104 @@
...
+        """Indicates if the current edit state differs from the original loaded state.
+
+        Returns:
+            Returns True if the current edit state differs from the orginal
+            loaded state; otherwise False.
+        """
+        fields: list[EditorField] = self.get_value("fields")
+        api_fields: list[EditorField] = self.get_value("_api_fields")
+
+        return not _are_fields_equal_editwise(fields, api_fields)
```
cutoffthetop: could you explain why a simple `fields != api_fields` didn't work?

mex/mex.py
```diff
@@ -52,6 +52,18 @@
...
     ],
 )
+app.add_page(
+    create_index,
+    route="/create/[draft_identifier]",
+    title="MEx Editor | Create",
+    on_load=[
+        State.check_mex_login,
+        State.load_nav,
+        # CreateState.reset_stem_type,
```
cutoffthetop: ?

mex/editor/rules/local_storage_mixin_state.py
```diff
@@ -0,0 +1,88 @@
...
+    UserEdit,
+)
+from mex.editor.transform import transform_fields_to_title
+
+
+class LocalStorageMixinState(rx.State, mixin=True):
+    """State-Mixin for handling local drafts and edits of EditorFields."""
+
+    local_draft_store: str = rx.LocalStorage('{"value": {}}', sync=True)
+    local_edit_store: str = rx.LocalStorage('{"value": {}}', sync=True)
```
cutoffthetop, line 19: why does this need to be a mixin when it is used only once in the RuleState? neither the RuleState nor this mixin would work on its own: the RuleState has a dependency to the Mixin because `update_local_state` wants to call `self.update_draft()`. and the Mixin has a dependency on the business-logic level to the RuleState, because it can only handle UserDraft and UserEdit models. (so it couldn't easily be used by the ingest page for example.)

mex/editor/rules/models.py
```diff
@@ -50,3 +52,47 @@ class FieldTranslation(rx.Base):
...
+
+    identifier: str
+
+
+class UserDraft(LocalDraft):
+    """Model to represent local drafts."""
+
+    identifier: str
+    title: EditorValue
+
```
cutoffthetop, line 98: would it hurt to store the full UserDraft/UserEdit? would make this class hierarchy a bit flatter and easier to grasp

mex/editor/state.py
```diff
@@ -77,38 +75,6 @@ def change_locale(self, locale: str) -> None:
...
-        """
-        self.current_page_has_changes = value
-
-    @rx.event
-    def close_navigate_dialog(self) -> None:
-        """Close the navigate dialog."""
-        self.navigate_dialog_open = False
-
-    @rx.event
-    def navigate(self, raw_path: str) -> EventSpec | None:
```
cutoffthetop, line 95: 👋🙂

CHANGELOG.md
```diff
@@ -9,14 +9,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
 ### Changes
 
 - updated template to https://github.com/robert-koch-institut/mex-template/commit/7c3e75
+- Using href navigation instead of navigation via code (`page.navigate()`).
+- Initial language now picking from available languages instead of hard coded value
+- bumped cookiecutter template to https://github.com/robert-koch-institut/mex-template/commit/3c389d
+- Increased global expect timeout (to 30,000) for tests
+- Using the field label for new/remove additive value button
+- renaming the field_name function to field_name_card
```
cutoffthetop: these are old changes, merging glitch?

cutoffthetop (changes requested):
nice work! the draft menu looks very clean 👍 i also like that we have proper hrefs in the navbar again :)
i have a few smaller suggestion and one discussion about the mixin.
back to you!

mex/editor/layout.py
eckelmannf: oh damn. Would be nice when ruff detects stuff like this o0

mex/editor/state.py
eckelmannf: I don't know how u could see this code since its not part of the commit. Even in this preview u can see the code is removed (- on the side). When I click on vscode on it. it opens and old file that is readonly 😕

mex/editor/rules/state.py
eckelmannf: because i check identifier-fields like this
```
state_value.identifier == api_value.identifier
if api_value.identifier and not api_value.text
else state_value.identifier == api_value.identifier and state_value.text == api_value.text 
```
the text is lazy loaded and thus it cant be used inside the check

mex/editor/rules/local_storage_mixin_state.py
eckelmannf: Yeah at this point i used the mixing to keep the boilerplate stuff away from the rulestate since its getting bigger and bigger. These funcs represent basicly the crud stuff for the local draft and edit manipulation. I really dont want to turm my monitor 90 degress even its the python meta :D

mex/editor/rules/models.py
eckelmannf: I like the fact we store the smalles piece of information on client side (localstorage max size 5mb). It also allows us to lazy load stuff from backend when resolving localstate to sessionstate. Seems more flexible and divided for me. I dont think that less classes/-hierarchies makes things easier. I like things explicit and clear.

mex/editor/layout.py
cutoffthetop: seems like they had it behind a feature-flag for a while and then paddled back 😞 https://github.com/astral-sh/ruff/pull/9463

mex/editor/rules/state.py
cutoffthetop: hm, i found compare_sequences and _are_fields_equal_editwise quite hard to follow. what do you think of using the object oriented approach here?

```python
from mex.common.types import MergedPrimarySourceIdentifier
from mex.editor.models import EditorValue as _EditorValue
from mex.editor.rules.models import EditorField as _EditorField
from mex.editor.rules.models import EditorPrimarySource as _EditorPrimarySource
from mex.editor.rules.models import InputConfig


class EditorValue(_EditorValue):
    def __eq__(self, other: object) -> bool:
        if isinstance(other, EditorValue):
            self_dict = self.dict(exclude={"text"} if self.identifier else {})
            other_dict = other.dict(exclude={"text"} if other.identifier else {})
            return self_dict == other_dict
        return super().__eq__(other)


class EditorPrimarySource(_EditorPrimarySource):
    def __eq__(self, other: object) -> bool:
        if isinstance(other, EditorPrimarySource):
            return (
                self.identifier == other.identifier
                and self.enabled == other.enabled
                and self.editor_values == other.editor_values
            )
        return super().__eq__(other)


class EditorField(_EditorField):
    def __eq__(self, other: object) -> bool:
        if isinstance(other, EditorField):
            return (
                self.name == other.name
                and self.primary_sources == other.primary_sources
            )
        return super().__eq__(other)


f1 = EditorField(
    name="foo",
    primary_sources=[
        EditorPrimarySource(
            identifier=MergedPrimarySourceIdentifier("0913801928301"),
            enabled=True,
            editor_values=[
                EditorValue(
                    text="Resolved Value",
                    identifier="9999129310293",
                ),
                EditorValue(
                    text="Static Value",
                ),
            ],
            name=EditorValue(
                text="Resolved Primary Source",
                identifier="0913801928301",
            ),
            input_config=InputConfig(),
        )
    ],
    is_required=True,
)


f2 = EditorField(
    name="foo",
    primary_sources=[
        EditorPrimarySource(
            identifier=MergedPrimarySourceIdentifier("0913801928301"),
            enabled=True,
            editor_values=[
                EditorValue(
                    identifier="9999129310293",
                ),
                EditorValue(
                    text="Static Value",
                ),
            ],
            name=EditorValue(
                identifier="0913801928301",
            ),
            input_config=InputConfig(),
        )
    ],
    is_required=True,
)

print("is_equal", f1 == f2)
```

cutoffthetop (commented):
just _are_fields_equal_editwise discussion open, the rest looks good

mex/editor/rules/state.py
eckelmannf: I think when it comes to equality there is not one single truth. it depends on the perspektiv what equality means. therefore i think its better to have multiple functions or classes that check for equality. but i think this is more a philosophy discussion than a programming discussion :D 
I likt the "dict exclude and compare them" approach alot. I would tend to have this as extra function. not only because i like the persepektive approach more when it comes to equality, but also im not sure if reflex is using the __eq__ built in for some special handling. when u want this inside the object i would still add another function like eq_editwise.

cutoffthetop: ok true, `__eq__` might be used by reflex for stuff we don't foresee. i still think the object-oriented approach is more readable here than the current combo of `_are_fields_equal_editwise` and `compare_sequences`

cutoffthetop: maybe something like this:

```python
from typing import Protocol
from mex.editor.models import EditorValue as _EditorValue
from mex.editor.rules.models import EditorField as _EditorField
from mex.editor.rules.models import EditorPrimarySource as _EditorPrimarySource


class ChangeDetector(Protocol):
    def has_changes(self, other: "ChangeDetector") -> bool: ...


def sequence_has_changes(left: ChangeDetector, right: ChangeDetector) -> bool:
    try:
        return any(a.has_changes(b) for a, b in zip(left, right, strict=True))
    except ValueError:
        return True  # sequences don't have same length


class EditorValue(_EditorValue):
    def has_changes(self, other: "EditorValue") -> bool:
        fields_to_exclude_on_self = {"text"} if self.identifier else set()
        self_dict = self.dict(exclude=fields_to_exclude_on_self)
        fields_to_exclude_on_other = {"text"} if other.identifier else set()
        other_dict = other.dict(fields_to_exclude_on_other)
        return self_dict != other_dict


class EditorPrimarySource(_EditorPrimarySource):
    def has_changes(self, other: "EditorPrimarySource") -> bool:
        return (
            self.identifier != other.identifier
            or self.enabled != other.enabled
            or sequence_has_changes(self.editor_values, other.editor_values)
        )


class EditorField(_EditorField):
    def has_changes(self, other: "EditorField") -> bool:
        return (
            self.name != other.name
            or sequence_has_changes(self.primary_sources, other.primary_sources)
        )
```

eckelmannf: Did u test this or is this just a code example? Because i wonder why u compare 2 empty dicts when there is no identifier. I also wonder why u dont sort the sequences before comparing/zipping. I think it would be more corrent when sorting the fields by their name before starting to compare, because we can't asume the fields are on the same order. When i have a table with sortable column i might want to reorder the fields and then this comparison becomes wrong.
I can adapt the oo-approach and put in the missing stuff if u ok with it.

cutoffthetop: no, i wrote this just as a dummy code example, free-hand.

- i am not comparing empty dicts, though. that was just part of the `exlude=` parameter value: `if you have an identifier, then exclude text, else don't exclude fields`
- field ordering is currently not possible in the UI, so i wouldn't think we need business logic for that here. following our YAGNI philosophy that would just be unnecessary complexity

eckelmannf: Ooops, didn't realize the closing bracket is behind the {}. I'll adapt accordingly.

mex/editor/create/state.py
```diff
@@ -11,15 +11,46 @@ class CreateState(RuleState):
     available_stem_types: list[str] = [r.stemType for r in RULE_SET_REQUEST_CLASSES]
 
     @rx.event
-    def set_stem_type(self, stem_type: str) -> None:
+    def set_stem_type(
+        self, stem_type: str
+    ) -> None:  # Generator[EventSpec, None, None]:
```
cutoffthetop:
```suggestion
    ) -> None:
```

mex/editor/utils.py
```diff
@@ -34,3 +37,50 @@ def load_settings() -> EditorSettings:
...
+
+    for i in range(left_len):
+        left_item = left[i]
+        right_item = right[i]
+        if key and key(left_item) != key(right_item):
+            return False
+        if not comparator(left_item, right_item):
+            return False
+
+    return True
```
cutoffthetop:
```suggestion
```

mex/editor/utils.py
```diff
@@ -1,3 +1,6 @@
+from collections.abc import Callable, Sequence
+from typing import Any, TypeVar
+
```
cutoffthetop:
```suggestion
```

cutoffthetop (dismissed):
👌🎅

PR #515, opened 2026-01-06
feature/mx-2009 inline reference search

eckelmannf:
### PR Context
Refactored the pagination into standalone state and component to avoid implementing it again (adapted states with pagination). Sorry for the big pr.

### Added
- Dialog to search for entities by name and entity-type.

### Changes
- Refactored pagination into state mixin and component with options.


mex/editor/rules/main.py
```diff
@@ -202,20 +203,36 @@ def textarea_input(
 
 
 def identifier_input(
-    field_name: str,
+    field_translation: FieldTranslation,
+    # field_name: str,
     index: int,
     identifier: str | None,
+    # reference_value_types: list[str],
```
cutoffthetop:
```suggestion
    index: int,
    identifier: str | None,
```

mex/editor/rules/transform.py
```diff
@@ -233,19 +234,30 @@ def transform_models_to_fields(
...
+        return result
+
     required_fields = get_required_mergeable_field_names(additive)
     fields_by_name = {
         field_name: EditorField(
             name=field_name,
+            value_type=[
+                _make_string_type(x, entity_type, field_name)
+                for x in ALL_TYPES_BY_FIELDS_BY_CLASS_NAMES[entity_type][field_name]
+            ],
```
cutoffthetop: why not just this?

```suggestion
            value_type = (
                REFERENCED_ENTITY_TYPES_BY_FIELD_BY_CLASS_NAME[entity_type][field_name],
            )
```

mex/editor/search/transform.py
```diff
@@ -18,3 +23,30 @@ def transform_models_to_results(
...
+            title=transform_models_to_title([model]),
+            preview=transform_models_to_preview([model]),
+            all_properties=transform_model_to_all_properties(model),
+            stem_type=model.stemType,
+        )
+        for model in models
+    ]
+
+
+def transform_model_to_all_properties(
```
cutoffthetop: could you de-duplicate this with `mex.editor.ingest.model_to_all_properties` please?

mex/editor/fields.py
```diff
@@ -41,3 +44,22 @@
     }
     for class_name, field_names in TEMPORAL_FIELDS_BY_CLASS_NAME.items()
 }
+
+# TODO(ND): Same as https://github.com/robert-koch-institut/mex-backend/blob/main/mex/backend/fields.py#L61
```
cutoffthetop:
```suggestion
# TODO(FE): Same as https://github.com/robert-koch-institut/mex-backend/blob/main/mex/backend/fields.py#L61
```

the (XY) is the author, not the assignee, so you can safely mention yourself here 😬

mex/editor/models.py
```diff
@@ -84,3 +88,72 @@ class ModelConfig(BaseModel):
...
+    ) -> "PaginationOptions":
+        """Create pagination options for a given state.
+
+        Args:
+            state: The state to create the options for.
+            on_page_change: EventHandler that gets executed when the current_page
+            changes. Defaults to None.
+
+        Returns:
+            _type_: _description_
```
cutoffthetop: ?

mex/editor/pagination_state_mixin.py
```diff
@@ -0,0 +1,75 @@
...
+class PaginationStateMixin(rx.State, mixin=True):
+    """State-Mixin for pagination behaviour."""
+
+    total: int = 0
+    limit: int = 50
+    current_page: int = 1
+
+    @rx.var
+    def max_page(self) -> int:
+        """Retrun the maximum page, based on total and limit."""
```
cutoffthetop:
```suggestion
        """Return the maximum page, based on total and limit."""
```

mex/editor/pagination_state_mixin.py
```diff
@@ -0,0 +1,75 @@
...
+
+    @rx.event
+    def set_total(self, total: int) -> None:
+        """Set the total of the pagination."""
+        self.total = total
+        self.set_current_page(self.current_page)  # type: ignore[misc]
+
+    @rx.event
+    def set_current_page(self, page_number: str | int) -> None:
+        """Set the current page (coerced to be between 1 aned max_page)."""
```
cutoffthetop:
```suggestion
        """Set the current page (coerced to be between 1 and max_page)."""
```

mex/editor/search_reference_dialog.py
```diff
@@ -0,0 +1,365 @@
...
+    def toggle_expand_properties(self, identifier: str) -> None:
+        """Toggle if all properties are visible for given identifier."""
+        if identifier in self.expanded_properties:
+            self.expanded_properties.remove(identifier)
+        else:
+            self.expanded_properties.append(identifier)
+
+    @rx.event
+    def cleanup_state_on_dialog_close(self, is_open: bool) -> None:  # noqa: FBT001
+        """Cleanup dialog state when closed to adress ComponentState bug."""
```
cutoffthetop:
```suggestion
        """Cleanup dialog state when closed to address ComponentState bug."""
```

mex/editor/search_reference_dialog.py
```diff
@@ -0,0 +1,365 @@
...
+def search_reference_dialog(
+    on_identifier_selected: EventType[str],
+    reference_types: rx.Var[Iterable[str]] | Iterable[str],
+    field_label: rx.Var[str] | str | None = None,
+) -> rx.Component:
+    """Create a button that opens a dialog to search for references."""
+    pagination_opts = PaginationOptions.create(
+        SearchReferenceDialogState, SearchReferenceDialogState.search
+    )
+    compid_prefix = "search-reference-dialog"
```
cutoffthetop: why not verbose? can save a few seconds for someone trying to figure out what a "compid" is 😅

```suggestion
    component_id_prefix = "search-reference-dialog"
```

mex/editor/pagination_state_mixin.py
```diff
@@ -0,0 +1,75 @@
...
+    def go_to_next_page(self) -> None:
+        """Navigate to the next page."""
+        self.set_current_page(self.current_page + 1)  # type: ignore[misc]
+
+    @rx.event
+    def reset_pagination(self) -> None:
+        """Reset the pagination to its default values."""
+        self.total = 0
+        self.current_page = 1
+        self.limit = 50
```
cutoffthetop: doesn't this conflict for SearchReferenceDialogState? that overwrites limit=10, so when you reset, it would go back to 50?

maybe we can use the built-in `State.reset()` method? or alternatively something like `PaginationStateMixin.get_fields()["limit"].default` to get the dynamic default

mex/editor/search_reference_dialog.py
```diff
@@ -0,0 +1,365 @@
...
+
+    limit = 10
+    user_query: str = ""
+    user_reference_types: list[str] = []
+
+    results: list[ReferenceDialogSearchResult] = []
+    expanded_properties: list[str] = []
+    is_loading: bool = False
+
+    _locale_service = LocaleService.get()
```
cutoffthetop, line 40: the locale service is a singleton anyway, right? and you can get the singleton instance via `LocaleService.get()`. so why not call that in the method were the service is needed directly? - like we do with the backend connector in the `search()` method

mex/editor/search_reference_dialog.py
```diff
@@ -0,0 +1,365 @@
...
+    reference_types: rx.Var[Iterable[str]] | Iterable[str],
+    field_label: rx.Var[str] | str | None = None,
+) -> rx.Component:
+    """Create a button that opens a dialog to search for references."""
+    pagination_opts = PaginationOptions.create(
+        SearchReferenceDialogState, SearchReferenceDialogState.search
+    )
+    compid_prefix = "search-reference-dialog"
+
+    def render_properties(props: list[EditorValue]) -> rx.Component:
```
cutoffthetop, line 164: is it possible to de-dupe this with `mex.editor.ingest.main.render_all_properties` ?

mex/editor/rules/models.py
```diff
@@ -52,6 +52,7 @@ class EditorField(rx.Base):
     """Model for describing the editor state for a single field."""
 
     name: str
+    value_type: list[str]
```
cutoffthetop, line 55:
```suggestion
    value_type: list[str] = []
```

can we set a default here? otherwise old draft items i had in my browser were not validating any more

mex/editor/search_reference_dialog.py
```diff
@@ -0,0 +1,365 @@
...
+            rx.cond(
+                SearchReferenceDialogState.results,
+                rx.vstack(
+                    rx.foreach(SearchReferenceDialogState.results, render_result_item),
+                    style=rx.Style(overflow="auto", max_height="50vh"),
+                    custom_attrs={"data-testid": f"{compid_prefix}-search-results"},
+                ),
+                rx.center(
+                    rx.text(
+                        SearchReferenceDialogState.label_results_none_found_format,
```
cutoffthetop, line 252: weirdly for me, it always shows the label_results_none_found_format message rather than the loading spinner. can you confirm that on your machine too?

cutoffthetop (changes requested):
the modal looks and behaves nicely!
thumbs up also for the unified pagination 👍
i have a few questions/ideas, nothing big though

mex/editor/rules/transform.py
eckelmannf: Because this are only the reference types. So types like str or int are not available. But i want it consistent over all editor fields, since the property is called valuetype. I would imply there is always a valuetype present for every field no matter if its str or specialReferenceTypeIdentifer.

mex/editor/search/transform.py
eckelmannf: Was also thinkin about it, but in the ingest page its stored inside the state (accesses via index) and i put it into my models. I can do it if u want me to, but it will involve some changes on the ingest page and im not sure if this leads to other unexpected behaviours or failures. If u want me to adapt the ingest page i would also put it inside the models there or considering using the same models (e.g. DetailedSearchResult or sth)

mex/editor/fields.py
eckelmannf: okay :)

mex/editor/models.py
eckelmannf: oops some leftovers from my docstring generator extention o0

mex/editor/search_reference_dialog.py
eckelmannf: Not the first time we discuss this :)
To get the Singleton there is logic needed (if not there instantiate) this logic is always called when i translate stuff. I can get rid of this by just get the singleton once. also this way the get can change and instatiate every time when its get called, but this way it doesnt matter because i only get it once for the whole state. i dont  see the probleme here? why it is better to say every time LocalService.get().translate() instead of getting it once and calling _localService.translate(). 
To sum up, I think my way is better because:
- i get rid of the extra calls (is singleton instantiated, otherwise...) 
- the get method can change its logic how its instantiate the service

mex/editor/pagination_state_mixin.py
eckelmannf: ure right good point. 
But then the dialog has to set the limit like `PaginationStateMixin.get_fields()["limit"].default = 10` ?

mex/editor/search_reference_dialog.py
eckelmannf: Mmh... they look slightly different to me. But it might be usefull to create something like a searchComponent that has the option to enable searchresult expandation. Since its the 3rd time we implement a search and all work pretty equal. IMO the Pagination is just a start  of this refactorisation. I dont think its usefull to only cherry pick some methods instead of reworking the whole thing.

mex/editor/rules/models.py
eckelmannf: Mmh... I think value_type should never be empty and this would be a wrong default. What about clearing your local browser storage (i assume u have an old and wrong storage then ^^)

mex/editor/search_reference_dialog.py
eckelmannf: No i can see the spinner when i click search. Its only there like 0.5s but it appears. But i missed a rx.center around the spinner :)

mex/editor/pagination_state_mixin.py
eckelmannf: NVM its working like u described ❤️

mex/editor/rules/transform.py
cutoffthetop: okay, i see. but could you move this to mex.editor.fields then and create a new lookup dictionary there? so that you can use it similarly here:

```
            value_type = (
                STRINGIFIED_TYPES_BY_FIELD_BY_CLASS_NAME[entity_type][field_name],
            )
```

that would be more consistent with how model/field/type inspection works (pre-computed into look-ups instead of inspecting during runtime)

mex/editor/search/transform.py
cutoffthetop: i am not sure i follow :thinking: i was just thinking that these two are exactly the same function, except for the more permissive typing on one of them. so they could be merged into one single function, eg in `mex/editor/transform.py`

```python
# mex/editor/ingest/transform.py

def model_to_all_properties(model: AnyExtractedModel) -> list[EditorValue]:
    """Transform all properties of a model into a list of EditorValues."""
    return [
        value
        for field_name in model.model_fields
        for value in transform_values(getattr(model, field_name), allow_link=False)
    ]

# mex/editor/search/transform.py

def transform_model_to_all_properties(
    model: AnyExtractedModel | AnyPreviewModel,
) -> list[EditorValue]:
    """Transform all properties of a model into a list of EditorValues."""
    return [
        value
        for field_name in model.model_fields
        for value in transform_values(getattr(model, field_name), allow_link=False)
    ]
```

mex/editor/search_reference_dialog.py
cutoffthetop: i was mainly after consistency with the backend connector here, but i am not too emotionally invested here so we can just leave it as is^^

mex/editor/search_reference_dialog.py
cutoffthetop: fair enough

mex/editor/rules/transform.py
eckelmannf: right! good catch

mex/editor/search/transform.py
eckelmannf: oh i missunderstood u there.  U're absolutly right. Thought u want me to unify the whole expandedProperties thing.

mex/editor/search_reference_dialog.py
eckelmannf: tbh i did this because otherwise my string formats exceed the line limit ^^

tests/ingest/test_transform.py
```diff
@@ -16,10 +16,10 @@ def test_model_to_all_properties() -> None:
     model.model_fields = {"field1": Mock(), "field2": Mock()}
 
     with patch(
-        "mex.editor.transform.transform_values",
+        "mex.editor.transform.transform_model_to_all_properties",
         side_effect=lambda x, allow_link: [EditorValue(text=f"value{x}")],
     ):
-        result = model_to_all_properties(model)
+        result = transform_model_to_all_properties(model)
```
cutoffthetop: since this test is now testing a function from mex.editor.transform, could you move this test to tests.test_transform?

cutoffthetop (changes requested):
just one small cleanup request, otherwise good to go

cutoffthetop (approved):
🚢

PR #517, opened 2026-01-06
feature/mx-2109 setup cve scanning for package lock json

MikaMk4:
### PR Context

refer to https://github.com/robert-koch-institut/mex-drop/pull/389

additionally, we need to spin up a backend container because for some reason reflex tries to connect to it even when just running `reflex export`

### Added
- add cve scanning of frontend dependencies

### DIFF
--- a/.github/workflows/cve-scan.yml
+++ b/.github/workflows/cve-scan.yml
@@ -67,7 +67,7 @@ jobs:
           python-version: 3.11
 
       - name: Install requirements
-        run: make setup
+        run: make install
 
       - name: Export dependencies
         run: |
--- a/.github/workflows/cve-scan.yml
+++ b/.github/workflows/cve-scan.yml
@@ -21,6 +21,20 @@ jobs:
   scan:
     runs-on: ubuntu-latest
     timeout-minutes: 10
+
+    services:
+      backend:
+        image: ghcr.io/robert-koch-institut/mex-backend:1.2.0
+        env:
+          MEX_BACKEND_API_USER_DATABASE: ${{ secrets.MEX_BACKEND_API_USER_DATABASE }}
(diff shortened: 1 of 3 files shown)


.github/workflows/cve-scan.yml
```diff
@@ -21,6 +21,20 @@ jobs:
...
+        env:
+          MEX_BACKEND_API_USER_DATABASE: ${{ secrets.MEX_BACKEND_API_USER_DATABASE }}
+          MEX_BACKEND_API_KEY_DATABASE: ${{ secrets.MEX_BACKEND_API_KEY_DATABASE }}
+          MEX_IDENTITY_PROVIDER: graph
+          MEX_GRAPH_URL: neo4j://neo4j:7687
+          MEX_DEBUG: True
+        ports:
+          - 8080:8080
+        options: --entrypoint testing-backend
+
```
cutoffthetop: i found the fix, so that we don't need to start a backend. you need to set the initial value for the backend version, like this:

```diff
--- a/mex/editor/state.py
+++ b/mex/editor/state.py
@@ -153,7 +153,7 @@ class State(rx.State):
         """Return the version of mex-editor."""
         return version("mex-editor")
 
-    @rx.var(cache=True)
+    @rx.var(cache=True, initial_value="N/A")
     def backend_version(self) -> str:
         """Return the version of mex-backend."""
         connector = BackendApiConnector.get()
```

cutoffthetop (changes requested):

.github/workflows/cve-scan.yml
MikaMk4: 👍

cutoffthetop (approved):

cutoffthetop (dismissed):

PR #518, opened 2026-01-09
feature/mx-2096-wrap-up-consent-page

ZehraVictoria:
### Changes

- username in user profile
- filter results based on person
- add publications
- fix linking to MEx Web App
- readable timestamp
- color coded status
- fix and add sometranslations


mex/editor/layout.py
```diff
@@ -15,11 +15,12 @@
 locale_service = LocaleService.get()
 
 
-def user_button() -> rx.Component:
+def user_button(user_type: str = "user_mex") -> rx.Component:
```
eckelmannf: I think the arg should be `User | Var[User]`

mex/editor/layout.py
```diff
@@ -28,15 +29,16 @@ def user_button() -> rx.Component:
     )
 
 
-def user_menu() -> rx.Component:
+def user_menu(user_type: str = "user_mex") -> rx.Component:
```
eckelmannf: I think the arg should be User | Var[User] or Literal["user_mex", "user_ldap"] with user = getattr(State, arg)

eckelmannf (commented):
Very nice! Awesome that u tested the new transform method. 
I only have a small change request: we introduced a pagination state mixin that we wanna try to reuse for our pagination stuff. I created a [PR](https://github.com/robert-koch-institut/mex-editor/pull/559) that depends on your branch that tries to accomplish this with a ComponentState. Im not sure about some decisions i made there (see comment and pr text).

mex/editor/layout.py
ZehraVictoria: [63ebe6f](https://github.com/robert-koch-institut/mex-editor/pull/518/commits/63ebe6f8f8a02a00383d939deb8bdccca52d389e)

mex/editor/layout.py
ZehraVictoria: Angepasst in [63ebe6f](https://github.com/robert-koch-institut/mex-editor/pull/518/commits/63ebe6f8f8a02a00383d939deb8bdccca52d389e)

eckelmannf (dismissed):

cutoffthetop (dismissed):

PR #519, opened 2026-01-13
feature/mx-2156 search result component

eckelmannf:
### Added
- Component to display search results with different options


mex/editor/search_results_component.py
```diff
@@ -0,0 +1,221 @@
...
+    )
+
+
+@dataclass
+class SearchResultsComponentOptions:
+    """Options for the search results component."""
+
+    summary_text: rx.Var[str]
+    list: SearchResultsListOptions
+    pagination: PaginationOptions | None = None
```
cutoffthetop: why not use a field(default_factory=...) here as well, like you did with SearchResultsListOptions.item_options?

mex/editor/search_results_component.py
```diff
@@ -0,0 +1,221 @@
...
+        custom_attrs={"data-testid": "search-results-summary"},
+    )
+
+
+@dataclass
+class SearchResultsComponentOptions:
+    """Options for the search results component."""
+
+    summary_text: rx.Var[str]
+    list: SearchResultsListOptions
```
cutoffthetop: using names of python's built-in functions or types is discouraged, as it can lead to hard to debug errors

```suggestion
    list_options: SearchResultsListOptions
```

mex/editor/ingest/state.py
```diff
@@ -32,11 +32,11 @@ def current_results_length(self) -> int:
         return len(self.results_transformed)
 
     @rx.event
-    def toggle_show_properties(self, index: int) -> None:
-        """Toggle the show properties state."""
-        self.results_transformed[index].show_properties = not self.results_transformed[
+    def toggle_show_all_properties(self, item: IngestResult, index: int) -> None:
+        """Toggle if all properties are visible for given identifier."""
```
cutoffthetop: > for given identifier

is it given? i just see `item: IngestResult, index: int` in the signature

mex/editor/merge/main.py
```diff
@@ -211,25 +175,28 @@ def search_panel(category: Literal["merged", "extracted"]) -> rx.Component:
...
-            category == "merged",
-            rx.foreach(
-                MergeState.results_merged,
-                lambda result, index: search_result(result, index, category="merged"),
-            ),
-            rx.foreach(
-                MergeState.results_extracted,
-                lambda result, index: search_result(
-                    result, index, category="extracted"
+        rx.el.div(
```
cutoffthetop:
```suggestion
        rx.box(
```

in other places we are using rx.box instead of rx.el.div, because that looked like the reflex-way to do it. fine either way, should just be consistent.

```python
class Box(elements.Div, RadixThemesComponent):
    """A fundamental layout building block, based on `div` element."""

    tag = "Box"
```

mex/editor/ingest/main.py
```diff
@@ -181,20 +102,20 @@ def search_results() -> rx.Component:
...
+                pagination=build_pagination_options(
+                    IngestState, IngestState.flag_ingested_items
+                ),
+                list=SearchResultsListOptions(
+                    item_options=SearchResultsListItemOptions(
+                        enable_show_all_properties=True,
+                        on_toggle_show_all_properties=IngestState.toggle_show_all_properties,
+                        render_append_fn=ingest_button,  # type: ignore[arg-type]
+                    )
+                ),
```
cutoffthetop, line 118: nice cleaning up in the main files here 🧹

mex/editor/search/state.py
```diff
@@ -15,12 +15,13 @@
...
 from mex.editor.locale_service import LocaleService
-from mex.editor.pagination_state_mixin import PaginationStateMixin
+from mex.editor.models import SearchResult
+from mex.editor.pagination_component import PaginationStateMixin
 from mex.editor.search.models import (
     ReferenceFieldFilter,
     ReferenceFieldIdentifierFilter,
     SearchPrimarySource,
-    SearchResult,
+    # SearchResult,
```
cutoffthetop:
```suggestion
```

mex/editor/search_results_component.py
```diff
@@ -0,0 +1,221 @@
...
+
+    item_options: SearchResultsListItemOptions = field(
+        default_factory=SearchResultsListItemOptions
+    )
+
+
+def search_results_list(
+    items: list[SearchResult] | rx.Var[list[SearchResult]],
+    options: SearchResultsListOptions | None = None,
+    **kwargs: dict[str, Any],
```
cutoffthetop:
```suggestion
    style: rx.Style | None = None,
```

why not just accept Style? the rest of the kwargs are ignored anyway

mex/editor/search_results_component.py
```diff
@@ -0,0 +1,221 @@
...
+            else title,
+            render_additional_titles(item.title[1:]),
+        ]
+
+        if options.enable_show_all_properties ^ bool(
+            options.on_toggle_show_all_properties
+        ):
+            error_msg = "'enable_show_all_properties' is 'True' but "
+            "'on_toggle_show_all_properties' is not set; or vice versa."
+            raise ValueError(error_msg)
```
cutoffthetop: validating the config here feels a bit odd. wouldn't it make more sense to move the validation to the instantiation of SearchResultsListItemOptions?

either using [`__post_init__`](https://docs.python.org/3/library/dataclasses.html#dataclasses.__post_init__), by making the options pydantic models or by axing enable_show_all_properties because it can be derived from on_toggle_show_all_properties anyway

mex/editor/search_results_component.py
```diff
@@ -0,0 +1,221 @@
...
+            style=rx.Style(
+                color="var(--gray-12)",
+                fontWeight="var(--font-weight-light)",
+            ),
+            wrap="wrap",
+            align="center",
+            custom_attrs={"data-testid": f"display-properties-{property_type}"},
+        )
+
+    def search_results_item(
```
cutoffthetop: could you pull render_properties and search_results_item out of search_results_list please? there is no need for them to be nested, and it just makes search_results_list harder to read

cutoffthetop (changes requested):
lgtm, just a few questions/ideas

mex/editor/search_results_component.py
eckelmannf: There is a need: I dont want anyone else to call this functions. These are internal functions only used by search_result_list. I dont want to expose them.

mex/editor/search_results_component.py
eckelmannf: I dont agree on this. Values can change after init and thus can only be validated before using them.

mex/editor/search_results_component.py
eckelmannf: Because this is diffrent. Here `None` is absolutly fine, because its a good default (no pagination). On the otherside the default is `SearchResultsListItemOptions()` and not `None`.

mex/editor/search_results_component.py
cutoffthetop: hm, how about https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.validate_assignment though? that would guard against other functions using SearchResultsListItemOptions and forgetting to check the options integrity:

```python
from typing import Self
from pydantic import BaseModel, model_validator

class SearchResultsListItemOptions(BaseModel, validate_assignment=True):
    """Options for rendering a search results list item."""

    enable_show_all_properties: bool = False
    on_toggle_show_all_properties: EventCallback[SearchResult, int] | None = None
    ...

    @model_validator(mode="after")
    def validate_show_all_properties(self) -> Self:
        """"Show all properties can only be enabled when you provide a callback."""
        if self.enable_show_all_properties ^ bool(
            self.on_toggle_show_all_properties
        ):
            error_msg = "'enable_show_all_properties' is 'True' but "
            "'on_toggle_show_all_properties' is not set; or vice versa."
            raise ValueError(error_msg)
        return self
```

mex/editor/search_results_component.py
cutoffthetop: i don't see the harm in exposing them. i just find functions with 100+ lines hard to read, plus with the nested functions, it's quite tricky to figure out which variable names are coming from the signature of the nested function, and which are coming from the signature (or code) from the outer function. but i admit, it's somewhat a matter of taste.
as a compromise, how about making them `_private`? i know, that's not runtime-enforced in python but at least you'd get a ruff error when trying to use it elsewhere

eckelmannf: Seems like we are not reading code the same way. I stil thin its pretty ugly to expose functions because its more readable. But as u wish, i make them private.

mex/editor/search_results_component.py
eckelmannf: I think u missunderstood this ValueError. It was more like a hint for devs to when compiling the reflex app. A print warning would be enough since the error is raised due to reflex compilation anyway. The way u implemented it would lead to this strange behaviour.
`
opt = SearchResultsListOptions()
opt.show_all_properties = True -> Error!
opt.on_toggle_show_all_properties = lambda...
`
When i used the component i missed the eventhandler at some point and i was wondering why. This check should help to avoid this. Didn't expect this to be a problem. Before using a validation that forces me to set properties in the correct order i would get rid of this error.

cutoffthetop (approved):
lgtm 🚀

PR #539, opened 2026-01-26
feature/reflex 0.8.14

cutoffthetop:
### PR Context
- i chose this half-step upgrade, because reflex 0.8.15 deprecates `rx.BaseModel` which might have a deeper impact than the changes 0.7.14-0.8.14

### Changes
- upgrade reflex to 0.8.14
  - rework routing and nav items, because `State.router.page` is deprecated
  - rework search ref dialog, because the setTimeout script didn't work anymore
  - plus some smaller fixes and typing stuff


mex/editor/search_reference_dialog.py
```diff
@@ -66,18 +65,12 @@ def label_results_none_found_format(self) -> list[str]:
...
-        # TODO(FE): PLACEHOLDER - translate reference types when doin MX-2092
-        dummy_translation = self._locale_service.get_field_label(
-            self.current_locale, "", "contact"
-        )
-        result = [
-            f"{self.current_locale}: {x} ({dummy_translation})"
-            for x in self.user_reference_types
-        ]
-        return "|".join(result)
+        # TODO(FE): PLACEHOLDER - translate reference types when doing MX-2092
```
eckelmannf: Plz undo this changes since its already implemented in another ticket. It only will make merging painful.

mex/editor/search_reference_dialog.py
```diff
@@ -302,6 +273,12 @@ def render_search_form() -> rx.Component:
...
             style=rx.Style({"max-width": "62vw !important"}),
+            on_open_auto_focus=[
+                SearchReferenceDialogState.set_user_query(""),  # type: ignore[operator]
+                SearchReferenceDialogState.set_user_reference_types(reference_types),  # type: ignore[operator]
+                SearchReferenceDialogState.search,
+                SearchReferenceDialogState.resolve_identifiers,
+            ],
         ),
         on_open_change=SearchReferenceDialogState.cleanup_state_on_dialog_close,
     )
```
eckelmannf: No need for this. Reflex adds `set_<attribute>(value)` functions, to reduce boilerplate.

```suggestion

```

eckelmannf, line 295: Oh yeah thats a great way to do this. Much better, much cleaner. Thanks for reworking this <3

I just have two questions: 
1. Why u change the layout of the dialog? 
2. Do we want to call the "init" events also on (open_change and open=true)? The other name implies it only gets executed when it gets the focus?! Which i assume is always, but i dont know :/

mex/editor/pagination_component.py
```diff
@@ -50,10 +48,10 @@ def disable_next_page(self) -> bool:
         return self.current_page >= self.max_page
 
     @rx.event
-    def set_total(self, total: int) -> None:
+    def set_total(self, total: int) -> Generator[EventSpec]:
         """Set the total of the pagination."""
         self.total = total
-        self.set_current_page(self.current_page)  # type: ignore[operator]
+        yield type(self).set_current_page(self.current_page)  # type: ignore[operator]
```
eckelmannf: Plz get rid of the yield stuff since its not needed here and adapt on caller side.

mex/editor/pagination_component.py
```diff
@@ -67,22 +65,21 @@ def go_to_first_page(self) -> None:
         self.current_page = 1
 
     @rx.event
-    def go_to_previous_page(self) -> None:
+    def go_to_previous_page(self) -> Generator[EventSpec]:
         """Navigate to the previous page."""
-        self.set_current_page(self.current_page - 1)  # type: ignore[operator]
+        yield type(self).set_current_page(self.current_page - 1)  # type: ignore[operator]
```
eckelmannf: Plz get rid of the yield stuff since its not needed here and adapt on caller side.

mex/editor/pagination_component.py
```diff
@@ -67,22 +65,21 @@ def go_to_first_page(self) -> None:
...
 
     @rx.event
-    def go_to_next_page(self) -> None:
+    def go_to_next_page(self) -> Generator[EventSpec]:
         """Navigate to the next page."""
-        self.set_current_page(self.current_page + 1)  # type: ignore[operator]
+        yield type(self).set_current_page(self.current_page + 1)  # type: ignore[operator]
```
eckelmannf: Plz get rid of the yield stuff since its not needed here and adapt on caller side.

mex/editor/search/state.py
```diff
@@ -175,17 +176,20 @@ def current_results_length(self) -> int:
...
+    def load_search_params(self) -> Generator[EventSpec]:
         """Load url params into the state."""
         router: RouterData = self.get_value("router")
-        self.set_current_page(router.page.params.get("page", 1))  # type: ignore[operator]
-        self.query_string = router.page.params.get("q", "")
-        type_params = router.page.params.get("entityType", [])
+        parsed_url = urlparse(router.url)
+        params = parse_qs(parsed_url.query)
+        current_page = params["page"][0] if "page" in params else 1
+        yield type(self).set_current_page(current_page)  # type: ignore[operator]
```
eckelmannf: The adaption on the caller side.

```suggestion
        self.set_current_page(current_page)  # type: ignore[operator]
        yield None
```

mex/editor/layout.py
```diff
@@ -101,12 +101,12 @@ def nav_link(item: NavItem) -> rx.Component:
         underline=item.underline,  # type: ignore[arg-type]
         class_name="nav-item",
         custom_attrs={
-            "data-testid": f"nav-item-{item.path}",
+            "data-testid": f"nav-item-{item.route_ids}",
```
eckelmannf: I think this will break alot of tests. Maybe only use route_ids[0]?

mex/editor/main.py
```diff
@@ -77,9 +74,10 @@ def editor_frontend() -> None:  # pragma: no cover
 
 
 def main() -> None:  # pragma: no cover
-    """Start the editor api together with frontend."""
+    """Start the editor api for local development."""
     # Set environment variables.
     environment.REFLEX_USE_GRANIAN.set(False)
+    environment.REFLEX_HOT_RELOAD_EXCLUDE_PATHS.set([Path("tests")])
```
eckelmannf, line 80: lovely <3

mex/editor/state.py
```diff
@@ -30,28 +30,28 @@ class State(rx.State):
     _nav_items: list[NavItem] = [
         NavItem(
             title="layout.nav_bar.search_navitem",
-            path="/",
-            raw_path="/?page=1",
+            route_ids=["/", "/index"],
```
eckelmannf, line 34: I think "/index" should be removed, since its no actived/valid route (not routed in mex.py)

```suggestion
            route_ids=["/"],
```

tests/create/test_main.py
```diff
@@ -74,7 +74,9 @@ def test_create_page_updates_nav_bar(create_page: Page) -> None:
     expect(nav_item).to_contain_text(
         build_ui_label_regex("layout.nav_bar.create_navitem")
     )
-    expect(nav_item).to_have_class(re.compile("rt-underline-always"))
+    assert nav_item.evaluate(
```
eckelmannf: Thats wild :D I think the best approach here would be to set a class like .active on the nav-item (to indicate its the active nav-item) and then check this. The the css class decides how its represented in the ui (maybe they just want the nav-item to be bold or show an icon beside or ...)

tests/edit/test_main.py
```diff
@@ -38,11 +38,13 @@ def edit_page(
...
     page = edit_page
     nav_bar = page.get_by_test_id("nav-bar")
-    page.screenshot(path="tests_edit_test_main-test_edit_page_updates_nav_bar.png")
     expect(nav_bar).to_be_visible()
+    page.screenshot(path="tests_edit_test_main-test_edit_page_updates_nav_bar.png")
     nav_item = nav_bar.locator(".nav-item").all()[2]
-    expect(nav_item).to_have_attribute("data-testid", "nav-item-/item/[identifier]")
-    expect(nav_item).to_have_class(re.compile("rt-underline-always"))
+    expect(nav_item).to_have_attribute("data-testid", "nav-item-/item/[item_id]")
+    assert nav_item.evaluate(
```
eckelmannf: We should extract util or generalized func to check this (DRY)

eckelmannf (commented):
Great work. Really nice reworks for search_reference_dialog and nav-items (including params handling). We should make this update available asap since i feel like reflex speed up alot with update.

mex/editor/search_reference_dialog.py
```diff
@@ -93,6 +86,16 @@ def cleanup_state_on_dialog_close(self, is_open: bool) -> None:  # noqa: FBT001
             self.results = []
             self.reset_pagination()  # type: ignore[operator]
 
+    @rx.event
```
eckelmannf, line 95: No need for this. Reflex adds set_<attribute>(value) functions, to reduce boilerplate.

mex/editor/search_reference_dialog.py
```diff
@@ -66,18 +65,12 @@ def label_results_none_found_format(self) -> list[str]:
     def label_results_select_button(self) -> None:
         """Label for results.select_button."""
 
-    @rx.var
+    @rx.var(cache=False)
```
eckelmannf: Why add cache=False?

eckelmannf:
Little Note: In order to make it run i had to set `NODE_TLS_REJECT_UNAUTHORIZED=0`

eckelmannf (changes requested):
Great work. Really nice reworks for search_reference_dialog and nav-items (including params handling). Just a few comments and questions.

mex/editor/search_reference_dialog.py
cutoffthetop: cache is on by default and i was seeing the type labels not being updated :/

mex/editor/search_reference_dialog.py
cutoffthetop: sorry, fixed in 5169b93

mex/editor/layout.py
cutoffthetop: b6fb29e

tests/create/test_main.py
cutoffthetop: yeah you're right :joy: i wanted to get rid off this built-in reflex class "rt-underline-always", but the better way would be to have our own class and not let the model dictate the style anyway. fixed in 2502481

tests/edit/test_main.py
cutoffthetop: see above

mex/editor/state.py
cutoffthetop: the /index route surfaced from somewhere of the depths of reflex :sweat_smile: i was guessing they add that as a magic route

mex/editor/search_reference_dialog.py
cutoffthetop: 1. i was just trying to simplify it a bit, as i was getting lost in this

```
        rx.fragment(
            rx.form(
                rx.vstack(
                    rx.hstack(
                        rx.hstack(
                            rx.input(
```

if i accidentally removed any needed complexity or ui behaviour here, let me know and i can restore

2. do you mean the `on_open_change=SearchReferenceDialogState.cleanup_state_on_dialog_close` line?

sadly there is no plain `on_close` though :/

mex/editor/pagination_component.py
cutoffthetop: a84ddcd

mex/editor/pagination_component.py
cutoffthetop: a84ddcd

mex/editor/pagination_component.py
cutoffthetop: a84ddcd

mex/editor/search/state.py
cutoffthetop: a84ddcd

mex/editor/search_reference_dialog.py
cutoffthetop: reflex is planning to remove them in v0.9

mex/editor/search_reference_dialog.py
eckelmannf: 1. i invested some time to make sure it also looks okay when i resize the window. thats why there are alot of v and hstacks contained in each other. im not sure if u changed it or not, hard to tell without testing manual, so i would prefer changing it back.

2. I just came up with the on_mount event which looks like exactly what we need. We might also can use for cleanup 😱 

From Reflex docs:
> Reflex components have lifecycle events like on_mount and on_unmount that allow you to execute code at specific points in a component's existence. These events are crucial for initializing data, cleaning up resources, and creating dynamic user interfaces.

I tried to used on_mount at other points but realized it often gets called more than once (i assume due to some rx.cond that changes over time). As a workaround i use .debounce(500) on functions which ensures it only gets called once within 500ms.

mex/editor/state.py
eckelmannf: hahaha. couldnt see any magic happen :D

mex/editor/search_reference_dialog.py
eckelmannf: Good catch. then this is perfeclty fine 👍

mex/editor/search_reference_dialog.py
eckelmannf: I think there was another problem anyway. The difference between cache True and Flase is:
True: detect dependencies self.<stateAttr> and recalc when one of those is changing
False: always recompute when an event is called
 :man_teacher: I think false is most of the time the wrong choice except u do sth like date.now()

cutoffthetop: gone d765878

mex/editor/search_reference_dialog.py
cutoffthetop: RE 1: i did a rudimentary width fix so its usable with a small viewport 48f8d70 - and since you also could not spot which responsive functionality i took away, you kinda proved my point that this was unreadable with 7 levels nesting depth 😬

RE 2: in theory, on_mount/on_unmount would be nice, but i couldn't get it to work here :/ reflex crashed with some internal error with `recursivelyTraverseAndDoubleInvokeEffectsInDEV` 🤷 this is what i tried:

```diff
--- a/mex/editor/search_reference_dialog.py
+++ b/mex/editor/search_reference_dialog.py
@@ -84,13 +84,12 @@ class SearchReferenceDialogState(State, PaginationStateMixin):
         self.results[index].show_all_properties = not item.show_all_properties

     @rx.event
-    def cleanup_state_on_dialog_close(self, is_open: bool) -> None:  # noqa: FBT001
+    def cleanup_state_on_dialog_close(self) -> None:
         """Cleanup dialog state when closed to address ComponentState bug."""
-        if not is_open:
-            self.user_query = ""
-            self.user_reference_types = []
-            self.results = []
-            self.reset_pagination()  # type: ignore[operator]
+        self.user_query = ""
+        self.user_reference_types = []
+        self.results = []
+        self.reset_pagination()  # type: ignore[operator]

     @rx.event
     def set_user_query(self, value: str) -> None:
@@ -276,12 +275,12 @@ def search_reference_dialog(
             ),
             custom_focus_script(),
             style=rx.Style({"max-width": "62vw !important"}),
-            on_open_auto_focus=[
+            on_mount=[
                 SearchReferenceDialogState.set_user_query(""),  # type: ignore[operator]
                 SearchReferenceDialogState.set_user_reference_types(reference_types),  # type: ignore[operator]
                 SearchReferenceDialogState.search,
                 SearchReferenceDialogState.resolve_identifiers,
             ],
+            on_unmount=SearchReferenceDialogState.cleanup_state_on_dialog_close,
         ),
-        on_open_change=SearchReferenceDialogState.cleanup_state_on_dialog_close,
     )
```

eckelmannf: RE 1: I never said its unreadable (hstack for input + button, hstack for searchBox and pagination, vstack which u also have) i said i invested alot of time in it so it behaves "correctly". Im not sure what u want me to do here... most obvious thing is: pagination where on the same row and pagination was wrapping to next row when not enough space for it. I dont really want to explain every wrap and flex property :/
If u dont want to change it back im fine with it but i would suggest to not change such things in future otherwise it doesnt make any sense to invest time in this.

eckelmannf: RE 2: Damn, but thats why reflex is so frustrating... But at least we tried :/

cutoffthetop: RE 1: i mean i didn't want to undo your effort here, it was just the easiest (and laziest) way for me to get this code to a more graspable nugget-size. anyway, for a solution we can both live with: i have now kept your flex structure and split out the input-and-button part, which i think is already a lot more readable -> 9358a4d
d'accord?

RE 2: yeah :/ but i guess not worth investing more time atm

eckelmannf: RE 1: yeah much better, thanks. Just one last minor thing. U change rx.spacer() to rx.el.br(). Is the spacer no longer available? i think a `<div style="margin-left: auto; margin-right: auto"></div>` is more suitable than a break.

RE 2: No dont invest any more. We tried...

cutoffthetop: 1ed805a :)

eckelmannf (dismissed):
Wonderful! Reflex 0.8.14 we are coming 🚀

PR #549, opened 2026-02-02
feature/mx-2130 advanced search page

eckelmannf:
### Added
New page for advanced searching, including complex reference filter


.pdm-python
```diff
@@ -0,0 +1 @@
+D:/code/git/mex-editor/.venv/Scripts/python.exe
```
cutoffthetop: 🌫️

eckelmannf: removed

mex/editor/advanced_search/main.py
```diff
@@ -0,0 +1,262 @@
...
+                class_name="truncate",
+                style=rx.Style(align_self="end", max_width="60%", display="block"),
+                on_click=[
+                    AdvancedSearchState.remove_ref_filter(index),  # type: ignore[operator]
+                    AdvancedSearchState.search,
+                    AdvancedSearchState.resolve_identifiers,
+                ],
+                custom_attrs={"data-testid": f"ref-filter-{index}-remove"},
+            ),
+            gap="4px",
```
cutoffthetop:
```suggestion
            gap="var(--space-1)",
```

mex/editor/advanced_search/main.py
```diff
@@ -0,0 +1,262 @@
...
+        align="stretch",
+    )
+
+
+def filter_references() -> rx.Component:
+    """Render filter references page."""
+    return rx.card(
+        rx.vstack(
+            rx.hstack(
+                rx.heading("Reference Filters", size="3"),
```
cutoffthetop: a few things still untranslated here, but i'm fine to move those to a follow-up

mex/editor/locales/de.po
```diff
@@ -212,6 +212,21 @@ msgstr "Leider wurden keine Ergebnisse für '{}' gefunden :("
 msgid "search_reference_dialog.results.select_button"
 msgstr "Auswählen"
 
+msgid "layout.nav_bar.advanced_search_navitem"
+msgstr "Erweiterte Suche"
+
+msgid "advanced_search.reference_filter.add_value"
+msgstr "Hinzufügen"
```
cutoffthetop, line 219: `{} hinzufügen` -> "Kontakt hinzufügen"

would be more natural than the current label "Hinzufügen Kontakt"

mex/editor/advanced_search/main.py
```diff
@@ -0,0 +1,262 @@
...
+                rx.button(
+                    rx.icon("plus"),
+                    on_click=AdvancedSearchState.add_ref_filter(
+                        AdvancedSearchState.all_fields_for_entity_types[0].value  # type: ignore[attr-defined]
```
cutoffthetop, line 198: not sure what's causing it, but when i click this button, the whole sidebar expands in width, which looks quite odd

mex/editor/advanced_search/state.py
```diff
@@ -0,0 +1,276 @@
...
+            for item in group:
+                if field_key not in fields_with_all_ref_types_and_label:
+                    fields_with_all_ref_types_and_label[field_key] = set()
+                fields_with_all_ref_types_and_label[field_key].update(
+                    item["value_types"]
+                )
+
+        def _build_select_item(key: str, value: Iterable[str]) -> ValueLabelSelectItem:
+            [field, label] = key.split("::")
+            return ValueLabelSelectItem(label=label, value=f"{field}:{','.join(value)}")
```
cutoffthetop: hmm, i don't like the string-magic here, why not add the list of values to the SelectItem class as well?

mex/editor/advanced_search/state.py
```diff
@@ -0,0 +1,276 @@
...
+            for entity_type in selected_entity_types
+            for field in REFERENCE_FIELDS_BY_CLASS_NAME[
+                ensure_prefix(entity_type, "Extracted")
+            ]
+        ]
+
+        fields_with_all_ref_types_and_label: dict[str, set[str]] = {}
+        for k, g in groupby(fields_with_type, lambda x: x["field"]):
+            group = list(g)
+            field_key = f"{k}::{group[0]['label']}"
```
cutoffthetop: dict keys can be tuples in python (`{("foo", "bar"): 42}`), i think that would make this less fragile and nicer to read

mex/editor/advanced_search/state.py
```diff
@@ -0,0 +1,276 @@
...
+
+        Returns:
+            The fields for the selected entity types.
+        """
+        selected_entity_types = (
+            self.all_entity_types if len(self.entity_types) == 0 else self.entity_types
+        )
+
+        fields_with_type: list[dict[str, Any]] = [
+            {
```
cutoffthetop: could you make this a class or at least typedDict?

mex/editor/search_reference_dialog.py
```diff
@@ -142,7 +142,8 @@ def search(self) -> Generator[EventSpec | None]:
         else:
             self.is_loading = False
             self.results = transform_models_to_dialog_results(response.items)
-            yield SearchReferenceDialogState.set_total(response.total)  # type: ignore[operator]
+            yield SearchReferenceDialogState.set_total(response.total)  # type: ignore[operator]#
```
cutoffthetop:
```suggestion
            yield SearchReferenceDialogState.set_total(response.total)  # type: ignore[operator]
```

cutoffthetop (changes requested):
looks good, just a few comments

mex/editor/advanced_search/main.py
eckelmannf: Missed this one. No, lets try to make this PR contain everything.

mex/editor/advanced_search/state.py
eckelmannf: Unfortunatly not. I would love to make to but the reflex build in select item only allows string. I dont want to biuld my own select nor handle an extra dict where those values live , sync the dict and pick from there. I know its not the most beutiful but i see this as a custom serialization. we can also can consider calling something like json.dump() and then reparse the string into a class and use this, but i thought this might me simplier. sry for the long story.

mex/editor/advanced_search/state.py
eckelmannf: didnt know this. thx

mex/editor/locales/de.po
eckelmannf: would be nice but i couldnt make it work. The problem is, that its called inside a foreach render function https://github.com/robert-koch-institut/mex-editor/blob/feature/mx-2130-advanced-search-page/mex/editor/advanced_search/main.py#L174 and i couldt make format call or anything there to work. its possible somehow but i think its alot of work for one label :/

mex/editor/advanced_search/main.py
eckelmannf: mmh i think its kinda nice that the sidebar gets bigger when something inside "requests" more size and no other element inside the container has reserved any space. this is kinda complex imo but we can define a fix width to the sidebar if u think this looks better.

mex/editor/locales/de.po
cutoffthetop: fair enough, let's leave it

mex/editor/advanced_search/main.py
cutoffthetop: it seems to push out the result list to the side rather than shrinking it, so it feels more like a glitch than a feature. i'd rather give the sidebar more room to breathe by default

mex/editor/advanced_search/state.py
cutoffthetop: i see. i actually like the json.dump() idea, should be a small fix and feels less wobbly

eckelmannf: I also cleaned up the mess abit.

cutoffthetop (approved):
🚢

PR #555, opened 2026-02-05
Feature/mx-2141-dialog-on-logout-if-unsaved-changes

eckelmannf:
### Added
- Dialog that warns about unsaved changes (drafts and edits)
- Package pytest-retry because tests often fail on github


cutoffthetop (approved):
looks good and works 👍

PR #559, opened 2026-02-10
feature/mx-2096-wrap-up-consent-page PaginationMixin

eckelmannf:
### PR Context
We introduced a state mixin for pagination that makes the pagination logic reusable (or at least tries). This PR is a try to use this mixin within the consent page along the different categories.

### Added
- ConsentCategoryList: ComponentState to use the PageinationStateMixin to handle the different category lists

### Changes
- Removes the pagination und fetching logik from the consent page and moved it the new ConsentCategoryList

### DIFF
--- a/mex/editor/component_option_helper.py
+++ b/mex/editor/component_option_helper.py
@@ -6,6 +6,7 @@ from mex.editor.pagination_component import (
     PaginationButtonOptions,
     PaginationOptions,
     PaginationPageOptions,
+    PaginationStateMixin,
 )
 from mex.editor.search.state import SearchState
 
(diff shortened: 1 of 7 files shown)


eckelmannf:
Im not sure if the ConsentCategoryList should load the data on its own or the page should still fetch the data and pass it into the ConsentCategoryList :man_shrugging: I had trouble to make the page.on_load event starts the fetch_data, since its now part of the componentState, which is inialized and managed be reflex. I use the on_mount event here to start the search, but the event is called twice (i think due to some rx.cond that changes at some point) so i added .debounce().

PR #564, opened 2026-02-16
feature/reflex 0.8.15

cutoffthetop:
### Changes

- upgrade reflex to 0.8.15
- use pydantic v2 BaseModel instead of rx.Base


eckelmannf (commented):
nice. approved!

eckelmannf (approved):

PR #589, opened 2026-03-02
feature/mx-2089-superseded-by-edit-page

eckelmannf:
Tried something new: loaded list as computed_var based on item_id instead of using event. Used this approach since there is no real inheritance in states and it became difficult to add special field loading/creation in RuleState.refresh. So i decdided to go with a computed_var.  

### Added
- 'supersededBy' refs are shown on the edit page

### Changes
- transform_model_to_search_results moved to root und unified with other functions (adapted usages)


cutoffthetop (approved):
looks good and works 👍

PR #609, opened 2026-03-10
Feature/mx 2077 show final fields

vyvytranngoc:
### PR Context
show final fields in the mex editor in order for mdmgmt to be able to do quality checks of editor items.
One example is to check the mex-editor item to x-system mapping which requires the identifierInPrimarySource of the Extracted Items.

### Added
- shows final fields in the mex editor
- added manually an EditorField for identifierInPrimarySource in _transform_model_to_editor_primary_sources
- added into InputConfig allow_subtractive and allow_preventive

### Changes
- adpdated tests because identifierInPrimarySource is not in MERGEABLE_FIELDS


cutoffthetop (dismissed):
nice, looks good! could you just remove reflex from the requirements.txt, it's already in the toml.. plus a main-merge. then: ready to go!

PR #617, opened 2026-03-17
hotfix/mx-2215-description-single-line

eckelmannf:
### PR Context
Since there was no real AC i was a bit creative and decided that ill wrap text fields (instead of truncating them) that are configured to render as textarea in edit mode. I also reworked the layout using flex only.

### Changes
- Text rendering for textarea/-box fields
- Layout cleanup -> flex it


cutoffthetop (dismissed):
flex!

PR #619, opened 2026-03-19
feature/mx-2137 prohibit consent retraction

ZehraVictoria:
### Added

- denial button deactivates if positive consent was already given

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- forbid users to change their consent status from positive to negative
 - shows final fields in the mex editor
 - added manually an EditorField for identifierInPrimarySource in _transform_model_to_editor_primary_sources
 - added into InputConfig allow_subtractive and allow_preventive
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+
 - forbid users to change their consent status from positive to negative
 - shows final fields in the mex editor
 - added manually an EditorField for identifierInPrimarySource in _transform_model_to_editor_primary_sources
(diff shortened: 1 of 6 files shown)


mex/editor/consent/main.py
```diff
@@ -61,9 +61,34 @@ def consent_box() -> rx.Component:
                 rx.button(
                     ConsentState.label_consent_box_no_consent_button,
                     on_click=ConsentState.submit_rule_set("denial"),  # type: ignore[operator]
+                    disabled=rx.cond(
+                        ConsentState.consent_status,
+                        ConsentState.consent_status.preview[0].badge  # type: ignore [union-attr]
```
eckelmannf: I think `ConsentState.consent_status.preview[0].badge == "VALID_FOR_PROCESSING"` should be a computed var in the state where u can check if the preview has any items. Something like
```
@rx.var
def is_consent_valid_for_processing(self) -> bool:
    return True if self.constent_status and self.constent_status.preview and self.consent_status.preview[0].badge == "VALID_FOR_PROCESSING" else False
```

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- forbid users to change their consent status from positive to negative
```
eckelmannf, line 12:
```suggestion

- forbid users to change their consent status from positive to negative
```

mex/editor/consent/state.py
```diff
@@ -173,6 +173,10 @@ def label_consent_status_declined_format(self) -> list[str]:
         """Label for consent.consent_status.declined_format."""
         return [self.consent_datetime]
```
eckelmannf, line 190: I think `submit_rule_set` should check if the change is differnt from local state and only send the request if it would change something. I hope backend is checking this too, but it feels weird that u can press on accept mutliple times and the toast always pops up telling u its saved.

eckelmannf (commented):
Very nice. Just some ideas.

mex/editor/consent/state.py
ZehraVictoria: fair point [cfc1975](https://github.com/robert-koch-institut/mex-editor/pull/619/commits/cfc19759c94d714897881089b4a1aa7654758f27)

mex/editor/consent/main.py
ZehraVictoria: [cfc1975](https://github.com/robert-koch-institut/mex-editor/pull/619/commits/cfc19759c94d714897881089b4a1aa7654758f27)

eckelmannf (approved):

PR #659, opened 2026-06-08
feature/mx-2303-workflow-rule

eckelmannf:
### Changes
- updated mex.common to version 2.0

### Fixed
- fix layout on create page

### DIFF
--- a/mex/editor/create/main.py
+++ b/mex/editor/create/main.py
@@ -10,29 +10,34 @@ from mex.editor.rules.main import (
 )
 from mex.editor.rules.models import FieldTranslation
 from mex.editor.rules.state import RuleState
+from mex.editor.style_helper import add_component_style, flex1_style, flex3_style
 from mex.editor.value_label_select import value_label_select
+
+flex1_col_style = flex1_style | {"max_width": "25%"}
 
 
 def editor_field(field_translation: FieldTranslation) -> rx.Component:
     """Return a horizontal grid of cards for editing one field."""
     field = field_translation.field
     return rx.hstack(
-        field_name_card(field_translation),
+        add_component_style(field_name_card(field_translation), flex1_col_style),
         rx.vstack(
             rx.foreach(
                 field.primary_sources,
                 lambda primary_source: rx.hstack(
-                    editor_primary_source_stack(
-                        field_translation,
-                        primary_source,
+                    add_component_style(
+                        editor_primary_source_stack(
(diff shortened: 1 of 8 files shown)


mex/editor/create/main.py
```diff
@@ -10,29 +10,34 @@
...
                 lambda primary_source: rx.hstack(
-                    editor_primary_source_stack(
-                        field_translation,
-                        primary_source,
+                    add_component_style(
+                        editor_primary_source_stack(
+                            field_translation,
+                            primary_source,
+                        ),
+                        flex1_style,
```
esinsj, line 36: these lines differ from the seemingly similar function "editor_field" in mex/editor/create/main.py.
Or are they meant to be different?

mex/editor/create/main.py
```diff
@@ -10,29 +10,34 @@
 )
 from mex.editor.rules.models import FieldTranslation
 from mex.editor.rules.state import RuleState
+from mex.editor.style_helper import add_component_style, flex1_style, flex3_style
 from mex.editor.value_label_select import value_label_select
 
+flex1_col_style = flex1_style | {"max_width": "25%"}
```
esinsj: the same style is also defined in mex/editor/rules/main.py and therefore not DRY.
Idea: Move flex1_col_style to style_helper.py and import it.

mex/editor/create/main.py
```diff
@@ -125,7 +130,9 @@ def index() -> rx.Component:
                 editor_field,
             ),
             validation_errors(),
+            align="stretch",
             style=rx.Style(
+                flex="1",
                 width="100%",
```
esinsj, line 139: the ` width="100%",` is removed in line 35 (before) but kept here. Is that on purpose?

esinsj (commented):
some questions and suggestions

mex/editor/create/main.py
eckelmannf: yes. one should take all remaining space and the other should grow as much as possible.

mex/editor/create/main.py
eckelmannf: yes. Because the editor_field is display different in create mode and therefore another layout and processing is needed.

mex/editor/create/main.py
eckelmannf: good catch 👍

esinsj (approved):
looks good to me

PR #678, opened 2026-06-15
feature/mx-2134 publish targets edit page

eckelmannf:
### Changes
- replaced switch to toggle all primary sources with new workflow publish targets


pyproject.toml
```diff
@@ -55,6 +55,7 @@ context = 5
 show_error_codes = true
 strict = true
 plugins = ["pydantic.mypy"]
+python_version = "3.14"
```
cutoffthetop, line 58: why was this needed? if this wasn't reflex specific, please propose the change to https://github.com/robert-koch-institut/mex-template/blob/1d816db77df20388022165cab12f45821d0b9f6d/mex-%7B%7B%20cookiecutter.project_name%20%7D%7D/pyproject.toml#L36 as well

cutoffthetop (changes requested):
looks good and works! just one config question

pyproject.toml
eckelmannf: Couldnt commit without this line because mypy didn't recognize the label_var overrides with generic types. This was the fix i found. ill add to the template. good catch 👍

cutoffthetop (dismissed):

PR #745, opened 2026-07-24
feature/mx-2370 mex-admin rename

cutoffthetop:
### PR Context

- when this is merged, the repo will also be renamed

### Changes

- rename mex-editor to mex-admin (at least on the surface): package, logo and settings
- BREAKING: the environment variable prefix changed from `MEX_EDITOR_` to `MEX_ADMIN_`


esinsj (commented):
looks good to me. 
I still find some 
- `editor-toast`s, 
- `EditorPrimarySource`s, 
- `EditorField`s,  
- `EditorValue`s (including functions containing the pattern `editor_value`) and an 
- `NotImplementedError, match="cannot transform NoneType to editor value"`.

Should that be renamed, too? Or is this not specific to mex-admin and is also used in mex-editor-ng?

cutoffthetop:
> looks good to me. I still find some
> ...
> Should that be renamed, too? Or is this not specific to mex-admin and is also used in mex-editor-ng?

@esinsj, that's what i meant with "at least on the surface" :smile: i thought it is not worth refactoring all of those internals since mex-admin will be phased out anyway.

esinsj (approved):

